PR #32, opened 2026-06-10
Add bold list-label periods detection category (#31)

### LINKED ISSUE #31: List treatment
When creating lists, I've noticed LLMs will do something like this this:

- **Developer GTM.** I've run developer marketing and ecosystems at three developer-tooling companies. I can tell you what works, what burns money, and which motions don't survive contact with a real audience.
- **Content distribution.** Chain of Thought, LinkedIn, Substack — useful for credibility narratives and reaching technical buyers.
- **Intros.** Years of conferences, podcast guests, and dev-tool operator network. Disproportionately strong in inference, evals, observability, and developer productivity.

Specifically, when creating this kind of list the LLM will go "Intros." then proceed to content, whereas most humans would type "Intros:". This minor punctuation difference is noticeable to me as an indicator of Ai writing in list form for bulleted lists.

conorbronsdon:
Adds a new detection category for a list-formatting tell from #31: LLMs end bold bullet-list labels with a period (`**Intros.**`, `**Content distribution.**`) where a person almost always uses a colon (`**Intros:**`). The colon introduces a gloss; the period reads as a sentence the following clause then continues past.

## Changes
- New `### Bold list-label periods` rule in `SKILL.md`, placed next to `### Inline-header lists` in the formatting area. Uses the issue author's before/after examples. Includes a carve-out for bold spans that are full standalone sentences.
- Catalog count goes from 48 to 49: README `pattern categories` bullet bumped, plus a representative table row.
- Version bumped 3.9.0 to 3.10.0 (SKILL.md frontmatter + plugin manifest), bundled plugin skill re-synced, dated CHANGELOG entry added.

## CI
Ran locally before pushing: `scripts/check-pattern-count.sh` (49 in sync), `npm test` (detector fixtures + CATEGORIES.md contract pass), plugin sync script (clean), manifest JSON valid.

Filed as a judgment rule (no detector `type`), consistent with other label/structure rules.

Closes #31


conorbronsdon:
This won't always be for bold list headings, if the list heading was non-bolded but used the . it would also be an indicator (though perhaps not quite as much of a slam dunk)

conorbronsdon:
Broadened per your comment: the category is now List-label periods and covers the unbolded form too. Bold labels stay the strongest signal; an unbolded short noun-phrase label ending in a period at the start of a bullet is flagged as the weaker variant. Tightened the carve-out so a short complete sentence opening a bullet is not flagged, only label-shaped fragments (1-4 words, no verb). Count guard and tests still green at 49.

PR #55, opened 2026-07-16
Add "load-bearing" to Tier 1 table

wilu222:
## Summary

Adds metaphorical "load-bearing" as a Tier 1 phrase, a recurring LLM tell (especially Claude) for labeling any dependency the argument rests on ("load-bearing assumption," "load-bearing claim," "load-bearing invariant"). Includes a construction carve-out so literal uses (`load-bearing wall`, `load-bearing beam`, etc.) are not flagged.

- Tier 1 row + carve-out in `SKILL.md` (v3.16.0)
- `TIER1_PHRASES` regex in `detector/patterns.js`
- Fixtures: metaphor fires `tier1`; construction nouns exempt
- Plugin skill synced; cursor rule ported

## Sources

- [Marek Šuppa — "Load-bearing" is becoming LLM speak](https://mareksuppa.com/til/load-bearing/)
- [Yaniv Bernstein (LinkedIn) — Opus 4.7 "AI slop" tell](https://www.linkedin.com/posts/ybernstein_opus-47-has-dropped-a-new-ai-slop-writing-activity-7452530977479774208-kbQA)
- [Developers Digest — How to Stop Claude from Saying 'Load-Bearing'](https://www.developersdigest.tech/blog/stop-claude-saying-load-bearing)
- [Hacker News discussion](https://news.ycombinator.com/item?id=48905248)
- [r/ClaudeAI — "That is load-bearing"](https://www.reddit.com/r/ClaudeAI/comments/1tob6q5/that_is_loadbearing/)

## Checklist

- [ ] If I added a detector `type`: it's documented in `detector/CATEGORIES.md` and has a fixture in `detector/patterns.test.js` (a true positive **and** a must-not-fire case)
- [ ] If I added a judgment-only rule: it's listed under "Skill-only" in `detector/CATEGORIES.md`
- [x] Any factual claim about how AI or humans write (e.g. "ChatGPT emits X", "humans rarely do Y") cites a source
- [x] The prose I added passes the skill's own audit (no AI-writing tells, terse bullets, no hollow intensifiers)
- [x] `CHANGELOG.md` entry added under a dated `## [X.Y.Z]` heading, and `SKILL.md` `version:` bumped if a rule changed


conorbronsdon:
Thanks for this — good catch on the phrase, and the sourcing is solid. I've pushed a follow-up commit (125ab07) rather than sending you round again; three things came out of review.

**The carve-out was firing on ordinary construction English.** It only exempted `load-bearing` immediately before one of eight bare nouns, so all of these flagged:

```
"the load-bearing capacity of the floor"
"a load-bearing structural wall"        (adjective in between)
"load-bearing masonry / partition / stud / lintel / pier / rafter / girder"
```

Added those nouns plus an optional material/position adjective slot. `structure`, `element`, `frame` and `foundation` are deliberately left *out* — they're used metaphorically too, and "the load-bearing structure of his argument" should still flag. There's a fixture pinning that.

**The `beam` fixture was passing vacuously.** `'The steel load-bearing beam spans 12 feet.'` is 7 words, and `analyzeText` returns early with zero issues when `wordCount < 10`, before any pattern runs. It passed with the carve-out deleted entirely. Mutation-testing showed only `walls?` was pinned by any assertion — the other seven nouns could all be dropped and the suite stayed green. Every fixture now clears the gate, asserts `!tooShort`, and each noun has its own case. Verified by mutation: deleting the carve-out, dropping just `capacity`, or dropping just the adjective slot each fail the suite now.

**`package.json` was still on 3.15.0** while SKILL.md, plugin.json and the cursor rule went to 3.16.0 — same drift class #51 fixed one release ago. CI only warns on this, so it would have slipped through.

Two notes on things you got right, since they cost you checklist ambiguity: reusing the `tier1` type means the `CATEGORIES.md` boxes correctly stay unchecked, and the plugin skill really is synced (both SKILL.md copies checksum identical). Also, if you wondered about the `filter: true` flag on `landscape`/`features` — ignore it, it's dead code, never read by the engine. Your inline lookahead was the right mechanism.

**One thing I'm not fixing here:** predicative use — "the wall in the kitchen **is** load-bearing" — still flags. That's not patchable with a lookahead, because the disambiguating word is the subject, which can sit arbitrarily far left. It's a design question about whether this carve-out should invert to an allowlist of *abstract* nouns, so I've opened #56 to gather feedback rather than block this PR on it.

Net: rule works, metaphor fires, literal attributive use is quiet. Merging with my follow up commit — shout if you disagree with any of it.

PR #61, opened 2026-07-24
Add moral-adjective category errors and invented contrast-pair mirror 

delight-f:
## Summary

Two new AI-slop detection rules discovered during a real audit (bacoff project docs/Explanation.md):

1. Moral-adjective category errors: catches AI gluing moral/character adjectives (honest, genuine, faithful) onto non-agentic technical nouns (shape, number, representation) where the adjective cannot literally modify the noun. Also covers passive-voice moral adverbs ('described honestly'), ontological slop on assumptions ('stops being true'), and gratuitous universal quantifiers.

2. Invented contrast-pair mirroring: catches AI fabricating the second half of a contrast pair for symmetry ('false precision rather than genuine accuracy') where one term is a real term of art and the other is phantom.

Both added to P1 severity tier, tolerance matrix (relaxed for technical-blog and docs), and skill-only section in CATEGORIES.md. Version bumped to 3.19.0.

## Checklist

- [ X] `npm test` passes (engine fixtures + `CATEGORIES.md` contract check)
- [ X] If I added a detector `type`: it's documented in `detector/CATEGORIES.md` and has a fixture in `detector/patterns.test.js` (a true positive **and** a must-not-fire case)
- [ X] If I added a judgment-only rule: it's listed under "Skill-only" in `detector/CATEGORIES.md`
- [ X] I considered false positives and added carve-outs for legitimate human writing
- [ X] Any factual claim about how AI or humans write (e.g. "ChatGPT emits X", "humans rarely do Y") cites a source
- [X ] The prose I added passes the skill's own audit (no AI-writing tells, terse bullets, no hollow intensifiers)
- [X ] `CHANGELOG.md` entry added under a dated `## [X.Y.Z]` heading, and `SKILL.md` `version:` bumped if a rule changed


conorbronsdon:
Nice catch on both of these. The moral-adjective category error and the invented contrast-pair are real tells, and "false precision rather than genuine accuracy" is a clean example of the phantom mirror. Merging these.

I pushed a few fixes to your branch to get it through CI:

- README pattern count was still 57; bumped it to 59 to match your two new categories. There's a CI check (`check-pattern-count.sh`) that enforces the README count against SKILL.md.
- `plugin.json` was on 3.18.0 while SKILL.md went to 3.19.0, so the plugin-sync check failed. Bumped it and regenerated the bundled skill copy.
- The changelog entry was dated 2025 and had picked up a lot of stray blank lines across the whole file, so I reset it and re-added a single clean 3.19.0 entry in the existing format.

Left your rule prose as you wrote it. `npm test`, `check-pattern-count.sh`, and `sync-plugin-skill.sh` are all green now. Thanks for the contribution.

conorbronsdon (approved):

PR #63, opened 2026-07-29
Add "verbatim" to the Tier 1 word table

mrlsecurity:
## What

Adds `verbatim` to the Tier 1 word table, with a terms-of-art carve-out.

## Why

The word is almost always redundant with the verb it modifies. "The packaging step copies the in-app resource verbatim" says nothing that "copies the in-app resource" doesn't: copying already implies the copy matches. So the rule is cut-first, and only when the exactness marks a contrast (an unmodified copy against a modified one) does it get a concrete replacement:

| context | replacement |
|---|---|
| bytes, files, binaries | `byte-for-byte` |
| text, quotes, speech | `word for word` |
| general | `unchanged`, `as is`, `exactly` |

## On sourcing

Per CONTRIBUTING, being explicit about what this rests on: I have no frequency study showing LLMs emit `verbatim` more often than humans do, and I am not claiming one. The argument is redundancy, which holds regardless of who wrote the sentence, plus the observation that it turns up in generated technical prose where the plain verb would do. I hit it in a bug report I had drafted with an LLM, where "copies the in-app resource verbatim" was flagged by a human reader as reading machine-written.

If you would rather see a frequency claim before taking a Tier 1 entry, Tier 3 (flag by density) is a reasonable home for it instead, and I am happy to move it.

## Precision over recall

`verbatim` has a genuine term-of-art sense, so this is a `TIER1_PHRASES` entry with a lookahead rather than a `TIER1` dictionary key. Exempted: `verbatim transcript`, `record`, `quote`/`quotes`/`quotation(s)`, `minutes`, `testimony`, `note`/`notes`, `reporter`/`reporters`, `reporting`. In legal, research and QA registers those are the standard name for the artifact, not an intensifier. The plural noun `verbatims` (market research) is exempt for free, since `\b` blocks the match.

## Tests

19 fixtures: 4 redundant uses that must fire, 15 carve-outs that must not, one per noun in the lookahead, so dropping any single noun fails the test. All padded past the `wordCount < 10` gate, following the `load-bearing` fixtures.

```
npm test                        # exit 0
scripts/check-pattern-count.sh  # pattern count in sync: 57
                                # word-table count in sync: 112 (60 + 40 + 12)
```

Reuses the existing `tier1` type, so `CATEGORIES.md` needs no new row and the contract check still holds. README word-table count went 111 to 112 to keep the guard script quiet.

## Left alone

No version bump and no CHANGELOG entry, since that is your call on release timing. Note `package.json` is still on 3.16.0 while `SKILL.md` and `plugin.json` are on 3.18.0; I did not touch it, but the drift predates this branch.


conorbronsdon:
Thanks for this, and for calling out the sourcing honestly up front. That made the call easy, and the lookahead carve-out was nicely done.

I moved it to **Tier 3** rather than merging at Tier 1. The argument here is redundancy, which holds for any writer, not that "verbatim" shows up more in AI text than human text. Tier 1 always fires, and "verbatim" is common in ordinary human writing, so an always-flag entry would catch people who aren't doing anything wrong. That's the false positive this repo works hardest to avoid. Tier 3 is density-gated, so it only fires on overuse, which is what the redundancy case actually supports.

In the detector, I dropped the `TIER1_PHRASES` regex and added `verbatim` to the flat `TIER3` list. That drops your terms-of-art lookahead, but density gating makes it unnecessary: a lone "verbatim transcript" never fires, only repetition does. I kept it out of `TIER3_PHRASES` since those entries feed the crypto/web3 cluster counter, and "verbatim" isn't that. I rewrote your test for the density behavior and pushed it to the branch. Counts hold at 112 (59 + 40 + 13); `npm test` and `check-pattern-count.sh` are green.

Keeping the branch name as-is. Thanks again for a clean, well-argued PR.

conorbronsdon (approved):

Issue #67, opened 2026-07-31
Em-dash carve-out misses changelog headings and bold-lead parentheticals


conorbronsdon:
Surfaced by `scripts/self-scan.js` (see `PROOF.md`), not by a user report.

`SKILL.md` carves out the list-item separator dash as typography rather than prose punctuation:

> an em dash acting as the separator in a bulleted or numbered list item that opens with a bolded lead term or a markdown link (`- **Term** — description`, `- [label](url) — description`) is typography, not a prose splice — don't count it toward the rate.

`detector/patterns.js` implements that with `SEPARATOR_DASH_RE`. Two near-identical shapes fall outside it.

**1. Keep-a-Changelog version headings.**

```
## [3.21.0] — 2026-07-30
```

Same job as the list-item separator: it joins a label to a value. `CHANGELOG.md` has 32 of these and every one counts toward the em-dash rate.

**2. A bold lead term with a parenthetical before the dash.**

```
- **Lingering-attention claims** (`lingering-attention`) — the share-post frame that claims...
```

`SEPARATOR_DASH_RE` requires the dash to follow the closing `**` with only whitespace between, so the parenthetical breaks the match.

**Measured on `CHANGELOG.md` at v3.22.0:** 137 em dashes total, 53 carved out, 84 counted. 33 of the 84 are the two shapes above. That is most of the file's residual score.

**Suggested fix:** extend the carve-out to (a) a heading line whose dash separates a bracketed or bolded label from a trailing value, and (b) an optional parenthetical or inline-code span between the bold lead term and the dash. Both should also be reflected in the `SKILL.md` rule text, since the carve-out is documented there first.

**Watch for:** the same regex feeds the smart-punct co-occurrence signature, so widening it lowers that signal too. Whatever lands needs a fixture proving a genuine mid-sentence prose splice in a list item still fires.

conorbronsdon:
Already fixed in 3.22.0 — verified against the shipped detector rather than by reading the regexes.

Both shapes are handled:

- **Changelog version headings** — `VERSION_HEADING_DASH_RE` was added and wired into *both* consumers, the em-dash rate and the smart-punct co-occurrence signature. It is deliberately narrow: bracketed or bare semver, dash, ISO date, nothing else on the line, so an ordinary prose dash in a heading still counts.
- **Bold lead with a parenthetical** — `SEPARATOR_DASH_RE` now allows an optional parenthetical or inline-code span between the closing `**` and the dash.

[...]

PR #68, opened 2026-07-31
Enforcement, proof, and the first false-positive measurement (v3.22.0)

conorbronsdon:
Closes the first two Tier 1 items from conorbronsdon/personal-context#119, which came out of reading the competitor repos in [Daguilar0123/ai-writing-skill-field-guide](https://github.com/Daguilar0123/ai-writing-skill-field-guide) directly.

## Preservation validator (`detector/validate.js`)

`SKILL.md` promises that a rewrite leaves certain things alone. Edit mode says "don't edit quoted material, code blocks, or text attributed to someone else"; rewrite mode says "preserve the original structure, intent, and all specific technical details." Nothing checked either promise, and edit mode writes to files, so a violation was silent and destructive.

`validate(original, rewritten)` errors on:

| code | what it caught |
|---|---|
| `code-block-modified` / `code-block-count` | fenced code edited, dropped, or added |
| `frontmatter-modified` | YAML frontmatter touched |
| `blockquote-modified` | someone else's quoted words reworded |
| `table-modified` | table cell content changed |
| `inline-code-missing` | an identifier, flag, or filename dropped |
| `url-missing` / `path-missing` | a link or path lost |
| `heading-count` / `heading-level` | the document restructured |
| `residual-grew` | the rewrite introduced more patterns than it removed |

Warnings cover reworded headings, figures that vanished from the original, and rewrites that drop more than 40% of the words.

Two carve-outs exist because a validator that fires on its own skill's instructions gets switched off within a day. URLs are compared with AI tracking parameters stripped from both sides, since the skill tells you to strip `utm_source=chatgpt.com`. Heading text changing is a warning rather than an error, since the skill tells you to sentence-case Title Case headings and cut emoji from them. Both cases have tests asserting they stay green.

Indented code blocks are counted but not enforced. Four-space indentation is also how markdown continues a list item, and blocking a legitimate rewrite is worse than reporting a soft signal.

There is a CLI: `node detector/validate.js before.md after.md`, exit 1 on any error. 23 tests, zero dependencies.

Prior art: `MohamedAbdallah-14/unslop`, `unslop/scripts/validate.py`. The carve-outs and the residual check are ours.

## Self-scan and PROOF.md

`isatimur/de-slop` scores its own docs with its own detector and publishes the result. Ours scored 67 on the README and 92 on SKILL.md, because a repo about AI writing quotes the patterns it warns about.

So `scripts/self-scan.js` publishes two numbers per file. **Raw** counts every match. **Exempt** applies the self-reference escape hatch that `SKILL.md` has documented in prose since v1 and never implemented: *"quoted examples are exempt from flagging. Text inside quotation marks, code blocks, or explicitly marked as illustrative should not be rewritten."* `applyExemptions()` is its executable form. SKILL.md goes 92 → 15, README 67 → 24.

Budgets gate the exempt column in CI. They are regression ceilings set from measured values with headroom, not quality claims, and the script says they only move down.

**The scan found two things in our own work**, and `PROOF.md` reports them rather than patching them quietly:

1. The em-dash rule carves out list-item separators but not Keep-a-Changelog version headings (`## [3.21.0] — 2026-07-30`) or a bold lead term with a parenthetical before the dash. 33 of the 84 counted em dashes in `CHANGELOG.md` are those two shapes. Filed as #67.
2. Roughly half of `CHANGELOG.md`'s residual score is release notes enumerating the words each new rule catches, unquoted, which the exemption cannot reach.

`PROOF.md` is in the table too, and scores 15 for exactly the reason it describes.

## Verification

```
$ npm test
All detector fixtures passed.
all preservation tests passed          (23/23)

$ npm run self-scan:check
PASS — every scanned document is within its score budget

$ bash scripts/check-pattern-count.sh
pattern count in sync: 61
word-table count in sync: 112 (59 + 40 + 13)
```

No new detection categories: the catalog stays at 61 and the engine at 46 `type`s. CI runs the self-scan as a step in the existing `detector` job rather than a second job, since branch protection pins required checks by job name.


conorbronsdon:
## Update: the remaining two Tier 1 items are in this PR

Pushed as a second commit. Same version (3.22.0), no new detection categories, catalog still 61 and engine still 46 `type`s.

### "Never inject these" guardrails (`SKILL.md`, under Tone calibration)

The section directly above tells the model to put voice back on purpose. That instruction has a predictable failure mode, and it is the one Vora's stress test found in `blader/humanizer`: generic AI phrasing replaced by a recognizable *humanizer* voice. A new fingerprint, not the absence of one.

Seven additions are now out of bounds regardless of how the result scores: fake first person, manufactured stakes, forced contrarianism, performed candor, em-dash theatrics, staccato conversion, and invented specifics.

The governing test is provenance. Subtraction and sharpening are in scope; addition of stance, personality, or fact is not. Adapted from `isatimur/de-slop`'s guardrails, which state it plainly.

It lives with the rewrite instructions rather than in the pattern catalog on purpose, and the section says why: a first-person aside is not a flag when the author wrote it and is a failure when the tool inserted it. That difference is provenance, which no pattern can see.

### False-positive intake (`.github/ISSUE_TEMPLATE/false_positive.yml`)

The form collects what makes a report measurable rather than anecdotal:

- the shortest text that still fires
- which rule fired
- **register** (blog, docs, technical blog, academic, social, email, issue/PR, changelog, fiction)
- how the text was actually written, from fully human through AI-drafted-then-rewritten
- whether the text may become a public fixture

Register is the load-bearing field. Patina's published numbers show false-positive rates ranging from 4% on chat to 34% on technical how-to prose within one language, so a single aggregate number would hide the thing a user needs to know. This is the intake that makes the Tier 2 corpus work possible.

The `false-positive` label was created to match. `PROOF.md`'s closing note now links the intake rather than only admitting the gap, and the README points at both the validator and `PROOF.md`.

### Verification after both commits

```
$ npm test
All detector fixtures passed.
all preservation tests passed

$ npm run self-scan:check
PASS — every scanned document is within its score budget

$ bash scripts/check-pattern-count.sh
pattern count in sync: 61
word-table count in sync: 112 (59 + 40 + 13)
```

SKILL.md's exempt score improved from 15 to 13 with the new section in place.

conorbronsdon:
## Third commit: human-control corpus and the first FP measurement

Reviewable on its own (`60707c9`). Independent of the first two commits; it only imports `detector/patterns.js`.

### Why

This repo has asserted things about false positives since v1. The tiering exists "to reduce false positives on words that are fine in isolation." The tolerance matrix relaxes rules per register. `SKILL.md` opens with "signals, not proof." None of it had been measured, and no English prose skill in the 68-repo survey publishes an FP rate.

### Design

Every document in `corpus/` was written by a person, so every flag on it is a false positive by construction. No labelling step, no judge, no model in the loop. Ground truth is provenance.

**Hash-only.** `manifest.json` carries what a document is, where it came from, its license, its register, and the sha256 of the measured text. Text is fetched to a gitignored `cache/` or read from wherever it already lives. Nothing gets republished, including the author's own writing. `verify` fails loudly on drift instead of re-recording, since a source that changed invalidates the measurement it backs.

**Register is the unit of analysis.** Patina measured 4% FP on chat prose and 34% on technical how-to inside a single language. An aggregate would be set entirely by the worst register.

### First result

303 paragraphs, nine public-domain works, 1788 to 1907.

| Threshold | n | flagged | FP rate | Wilson 95% CI |
|---|---:|---:|---:|---|
| score >= 25 | 303 | 0 | 0.0% | 0.0–1.3% |
| score >= 40 | 303 | 0 | 0.0% | 0.0–1.3% |
| score >= 50 | 303 | 0 | 0.0% | 0.0–1.3% |
| score >= 65 | 303 | 0 | 0.0% | 0.0–1.3% |

No paragraph of canonical human prose scored above **8 out of 100**.

**That is not a performance claim and `corpus/README.md` says so.** In patina's phrasing it is `no_calibration_signal_yet`: the corpus is too easy to expose a trade-off. Nobody runs this tool over *Walden*. They run it over READMEs, release notes, and LinkedIn posts, and none of those are in the corpus yet. Read the zero as a floor.

### What fired anyway, and the two defects it found

Document scores stayed at zero, but categories fired, and in `detect` mode every one of those reaches a user as a flag.

| Category | Units | Reading |
|---|---:|---|
| `em-dash` | 57 / 303 | corpus artifact: real 19th-century usage plus Gutenberg's `--` convention |
| `tier1` | 16 / 303 | real: `endeavor` ×5, `in order to` ×3, `commence` ×2, `ascertain` ×2 |
| `hollow-intensifier` | 6 / 303 | `truly` ×5 |
| `transition` | 5 / 303 | `moreover` ×3, `furthermore` ×2 |
| `hedge-stack` | 4 / 303 | three of four are a bug |

1. **#69**: `hedge-stack` allows two words between modal and hedge adverb, so `could not possibly` fires. Ordinary negation, and common.
2. **Nineteenth-century text cannot test the em-dash rule at all.** Era plus transcription convention. Recorded so nobody later reads 18.8% as a real rate.

Also worth a decision rather than an omission: `endeavor`, `commence`, and `ascertain` sit on the always-flag tier. In LLM output they are inflation; in legal, academic, and formal registers they are the ordinary word. That is the shape of a register-scoped relaxation the tolerance matrix does not currently have.

### No true-positive rate

The corpus contains no machine-written text. An FP rate alone says how often the tool cries wolf, not whether it can see wolves. `corpus/README.md` adapts patina's public-claim gate: no number here goes into the README or a social post until it covers registers people write in today.

```bash
npm run corpus          # list
npm run corpus:fetch    # populate cache
npm run corpus:verify   # hashes still match
npm run fp              # the measurement
```

conorbronsdon:
## Fourth commit: the corpus now has the maintainer's own pre-2023 writing, and the result changed

`22abec4`. Pulled 34 posts from the pre-2023 list on conorbronsdon.com, 2019 through October 2022, published before ChatGPT and in the register this tool is actually pointed at. `corpus.js` gained an HTML extraction path; text still never enters the repo, only its hash.

### Headline

**628 paragraphs. 0.0% false positives at every threshold. Wilson 95% CI 0.0–0.6%. Worst human paragraph: 11 out of 100.**

That now means something. The previous run could only say the detector doesn't fire on Victorian prose.

| Register | n | >=25 | >=40 | >=50 | >=65 |
|---|---:|---:|---:|---:|---:|
| blog | 351 | 0.0% | 0.0% | 0.0% | 0.0% |
| essay-literary | 141 | 0.0% | 0.0% | 0.0% | 0.0% |
| academic | 116 | 0.0% | 0.0% | 0.0% | 0.0% |
| technical-blog | 20 | 0.0% | 0.0% | 0.0% | 0.0% |

### The finding that matters more

`detect` mode does not show a document score. It shows every flag. On the maintainer's own pre-LLM posts (325 paragraphs, 2019–2022):

| Category | Paragraphs | Rate |
|---|---:|---|
| `em-dash` | 56 | **17.2%** |
| `tier1` | 38 | **11.7%** |
| `transition` | 9 | 2.8% |
| `uniformity` | 8 | 2.5% |

The Tier 1 words: `embrace` (9), `leverage` and inflections (8), `in order to` (5), `when it comes to` (4), `that said` (4), `features` (3), `thriving` (2), `robust` (2), `truly` (2), plus `utilize`, `pivotal`, `serves as`, `genuinely`, `in an era where`.

**Those are not AI tells in that text.** They are one writer's ordinary 2019 vocabulary, written years before the models existed. Tier 1 is documented as "always flag — replace on sight," and it fires on roughly one paragraph in eight of genuine pre-LLM professional writing.

`em-dash` is the same story and this time it is not a transcription artifact: one paragraph in six of his own pre-2023 prose exceeds the rate ceiling. The rule targets a habit he already had.

`corpus/README.md` records this as a decision rather than a defect, and states the counter-argument plainly: this is a writing-quality tool, and "`leverage` → `use`" is good advice whether or not a machine wrote the sentence.

### Corpus hygiene, both worth recording

- **Two guest posts** on the same site (Adam Noble, Steve Fawthrop) excluded by byline. Human-written, so they would not have corrupted the rate, but attributing them to the wrong author would have.
- **Three posts carrying "Looking back from 2025" retrospective inserts** dropped: their pre-LLM provenance is broken. Caught by reading the worst-scoring paragraphs, not by any check in the tooling.

Standing caveat now in the README: these are the *currently published* versions of pre-2023 posts, not verified originals. A silent post-2023 edit would not announce itself the way those three did. Fetching from the Wayback Machine at a pre-2023 timestamp is the fix and has not been done.

### Still not claimed

No true-positive rate. The corpus has no machine-written text. Registers `docs`, `social`, `email`, and `conversational` still have zero coverage.

```
$ npm test          → all passing
$ npm run fp        → the table above
$ npm run corpus:verify → 43/43 hashes match
```

conorbronsdon:
## Wayback switch: the corpus now reads verified pre-2023 originals

Commits `6c58bd8` (corpus helper tests) and `abf5750` (the switch).

The previous commit measured the *currently published* versions of posts written before 2023. That was the weakest claim in the corpus. It is now fixed.

### It mattered

The median archived capture is **0.92 similar** to its currently published counterpart, and five of 25 fall below 0.90. The live site's posts have been edited since. Measuring "pre-2023 writing" against pages edited in 2025 would have measured the wrong thing.

### Resolving the captures was not a domain swap

The old site used compressed slugs (`beveragetax`, `challengerfunnel`, `emailmarketing`) that don't match current URLs. So candidates were found by slug similarity and then **verified by content**: a capture is accepted only at 0.45+ Jaccard similarity against the current text.

Genuine matches land between 0.76 and 0.98. **Four slug guesses scored below 0.17 and were rejected**, which is the entire reason the check exists. `challengerfunnel` and `challengermarketing` were exactly the ambiguous pair the check was built for, and one of them failed.

**Nine posts dropped:** five have no archived snapshot, four failed the content check. Out rather than kept on current-site text, because a provenance rule that bends for convenience is not a provenance rule.

### Result

| Threshold | n | flagged | FP rate | Wilson 95% CI |
|---|---:|---:|---:|---|
| score >= 25 | 560 | 0 | 0.0% | 0.0–0.7% |
| score >= 40 | 560 | 0 | 0.0% | 0.0–0.7% |
| score >= 50 | 560 | 0 | 0.0% | 0.0–0.7% |
| score >= 65 | 560 | 0 | 0.0% | 0.0–0.7% |

The same measurement against current versions returned 0.0% across 628 paragraphs. **The headline does not depend on which copy was measured**, which is worth more than either number alone.

### The flag-level finding survives and strengthens

On the 25 Wayback-verified posts (257 paragraphs):

| Category | Verified originals | Current versions |
|---|---:|---:|
| `em-dash` | **18.3%** | 17.2% |
| `tier1` | **12.5%** | 11.7% |

Both rates are *higher* on the originals, which is what later editing passes would do. Tier 1 words: `embrace` (7), `leverage` and inflections (7), `when it comes to` (5), `in order to` (4), `that said` (4).

### Also in these commits

`scripts/corpus.test.js`, 15 tests on the extraction helpers, now in `npm test`. A silent extraction bug wouldn't crash anything; it would quietly change a published rate. Two tests exist purely to protect the measurement: em dashes must survive extraction, since the em-dash rule is scored against this corpus, and extraction must throw rather than return empty text, since an empty document would shrink the denominator without saying so.

```
$ npm test              → 0 failures
$ npm run corpus:verify → 34/34 hashes match
$ npm run fp            → the table above
```

conorbronsdon:
## Tier 1 split (`85ef3b4`)

Not a relaxation. Every word keeps the same edit advice; what changed is what a flag *claims*.

### The problem

Tier 1's definition is an empirical claim: these words *"appear 5–20x more often in AI text than human text. Replace on sight."* Nine members can't plausibly meet it — `in order to`, `utilize`, `serves as`, `features`, `boasts`, `presents`, `commence`, `ascertain`, `endeavor` are wordiness and formality edits. Good edits, filed under a tier whose definition asserts something else about them.

### The change

| | Tier 1A — frequency markers | Tier 1B — clarity edits |
|---|---|---|
| Examples | `delve`, `tapestry`, `beacon`, `testament to` | `in order to`, `utilize`, `commence` |
| Detector type | `tier1` | `tier1-clarity` (new) |
| Weight | 5 | 3 (same as Tier 2) |
| Counts toward dense-AI-vocab | yes | **no** |
| Edit advice | replace on sight | replace on sight |

A wordiness fix can no longer push a document toward an AI classification. Detect mode reports the bands separately.

### The measurement that prompted it

Against 257 paragraphs of verified pre-2023 human writing:

| | Before split | After |
|---|---:|---:|
| Tier 1 combined | 12.5% | — |
| 1A markers | — | **8.9%** |
| 1B clarity | — | **3.5%** |

No paragraph triggered both bands. **Roughly a quarter of Tier 1 hits on genuine human prose were wordiness being reported as an AI signal.** `commence` and `ascertain` firing on the Federalist Papers and Faraday is the same problem from the formal-register end.

### The 5–20x claim is now labelled

It traces to [`brandonwise/humanizer`](https://github.com/brandonwise/humanizer), credited in `SKILL.md` for the tiering. That repo asserts the ratio in two places and publishes **no method, dataset, or citation**. `SKILL.md` now says so, and commits to re-deriving the ratios once a machine-written corpus exists.

The wordiness conflation is inherited, not introduced here: `in order to`, `utilize`, `serves as`, and `robust` are all on the upstream list.

### Incidental fix

`check-pattern-count.sh` did not survive a tier containing two tables — its positional header detection never reset, so the second header counted as data (113 vs 112). Fixed, with a note in the file that its awk program is single-quoted and an apostrophe in a comment breaks the command substitution. Learned the hard way.

### Verification

```
$ npm test                        → 0 failures
$ bash scripts/check-pattern-count.sh
pattern count in sync: 61
word-table count in sync: 112 (59 + 40 + 13)
$ npm run self-scan:check         → PASS
```

Word-table total and catalog count both unchanged. The split also lowered this repo's own doc scores slightly (README 24 → 21, SKILL.md 13 → 12), since our docs quote the clarity words too — `PROOF.md` updated.

**Em dashes:** left alone deliberately. The maintainer has already cut back on them by choice, so the rule matches current practice; revisit later if that changes.

conorbronsdon:
## Machine-written corpus, and the first TPR this project has had (`27b6a04`)

Two external datasets, neither generated by anyone with a stake in the numbers.

| | RAID (Dugan et al. 2024, MIT) | HC3 (Guo et al. 2023, CC-BY-SA) |
|---|---|---|
| Task shape | in-domain continuation | assistant answers |
| Models | 11 families | ChatGPT, Dec 2022 |
| Design | domain-matched | **paired on question** |
| Units here | 373 machine, 37 human | 406 machine, 278 human |

`scripts/csv-lite.js` vendors a small RFC 4180 reader — RAID generations contain commas, quotes, and newlines, and splitting on delimiters would silently corrupt the text being measured.

## The result is negative and reported as such

| Threshold | FPR | TPR |
|---|---|---|
| >= 3 | 16.6% | 16.3% |
| >= 5 | 4.2% | 7.2% |
| >= 10 | 0.5% | 0.1% |
| >= 15 | 0.0% | 0.0% |

**ROC-AUC 0.501 paragraph level, 0.623 document level.** By source: HC3 0.654, RAID 0.599. Best operating point costs 12.8% FP to catch 27.7% of machine text.

The composite score does not separate the classes.

## Three findings underneath it

**1. The 0–100 scale uses a tenth of its range.** No paragraph of either class scored above 11. Every threshold ≥15 reads 0.0% on both sides, and `SKILL.md`'s own band files everything ≤15 as "Minimal AI signals" — so in practice every ordinary paragraph is labelled Minimal regardless of author. Weights run 2–12 and `rawScore` is divided by `max(1, log2(words/50))`. **Calibration defect, not detection failure, and the most fixable thing here.**

**2. The signal is structural, not lexical.** Document level:

| Category | Human | Machine | Lift |
|---|---:|---:|---:|
| `uniformity` | 2.1% | **25.1%** | **11.7x** |
| `filler` | 2.4% | 8.3% | 3.4x |
| `chatbot` / `hedge-stack` / `fnword-trigram-entropy` | 0.0% | ~1% | machine-only |
| `tier1` (112-entry table) | 8.0% | 7.4% | **0.9x** |

The vocabulary table — the thing the README leads with — fires slightly *more often on human writing*. This is exactly what `NulightJens/humanizer-stack` argues from StoryScope and `harshaneel/humanize` reaches independently. On this engine they look right.

**3. `em-dash` is inverted**: 9.9% human vs 1.9% machine, lift 0.2. On this corpus an em dash is evidence of *human* authorship. Unchanged as writing advice; recorded because it points the wrong way as evidence.

## Sampling note worth keeping

Evenly spaced byte offsets returned **fewer** model families at 40 windows than at 16 — they resonated with the file's domain-then-model sort order and missed gpt4, chatgpt, and cohere entirely. Offsets now follow a golden-ratio low-discrepancy sequence, and the builder warns at build time when model, domain, or unit coverage falls short.

## What this does not license

Not "the detector doesn't work". It measures how well one composite score separates two labelled classes on 2022–2024 corpora. The skill is documented as a writing-quality tool and none of this touches whether its edits improve prose. A lift near 1.0 says a category doesn't separate *these* classes on *these* corpora; `delve` is still worth replacing.

Every machine unit predates 2025, so these are an upper bound on current models, not an estimate for them. No local generation, deliberately: prompts written by someone who knows the pattern list would measure the prompting.

```
$ npm test              → 0 failures
$ npm run corpus:verify → 36/36
$ npm run fp            → tables above
```

conorbronsdon:
## Ninth commit: the two measured defects are fixed (`ecb9776`)

Both were filed off the corpus work, both now have fixtures, and both improved discrimination — which is the argument for measuring before patching.

**#69 `hedge-stack`** allowed two words between modal and hedge adverb, so `could not possibly` and `could a savage possibly` fired. Now at most one intervening word, never a negator.

**#67 em-dash carve-out** now covers Keep-a-Changelog version headings (`## [3.21.0] — 2026-07-30`) and a bold lead term with a parenthetical (`- **Term** (<code>slug</code>) — desc`). Deliberately narrow: a prose dash inside a heading still fires, with a fixture pinning that, because SKILL.md applies the rule to headings too.

Measured effect on the corpus:

| Category | Before | After |
|---|---|---|
| `hedge-stack` human rate | 0.6% | 0.2% |
| `hedge-stack` lift | <1 (fired more on human text) | **2.2** |
| `em-dash` human rate | 17.9% | 13.7% |

`em-dash` lift stays inverted at 0.1 — that is a finding about the signal, not a bug in the rule, and it is now tracked separately.

## Follow-through filed rather than silently dropped

Four issues, each with the measurement behind it and options rather than a foregone conclusion:

- **#70** — the 0–100 score uses a tenth of its range; nothing scores above 11, so every paragraph lands in "Minimal AI signals" regardless of author
- **#71** — the discriminating signal is structural (`uniformity` 11.7x) not lexical (vocabulary table 0.9x)
- **#72** — no post-2024 model in the corpus, which blocks both of the above
- **#73** — em-dash inverted as an authorship signal; decide its framing

#72 is the blocker on the other three. Re-weighting on 2022–2024 model habits would tune the engine to models nobody uses.

Issue #88, opened 2026-08-03
Decision: which style guides this repo can ship, and why there is no config-registry repo


conorbronsdon:
Opus 5 here: Recording a decision and the research behind it, so #76 and PR review threads can point here instead of re-arguing it.

## The question

#76 proposed a `--style` layer for house-style conformance. My first objection was that bundling CMOS/APA means tracking editions I can't test against. @ClaySecAI (and Conor) pushed back, correctly, that style guides move slowly. That left an open question: which style guides, if any, can this repo ship, and is a config library worth building as its own thing?

## What I found

**The config-registry idea is already built, by [Vale](https://github.com/vale-cli/vale).** Vale is a prose linter with a package registry, a GitHub Action, editor plugins, and a language server. Its registry `vale-cli/packages` was updated within the last week, and it already ships Vale-compatible implementations of the guides worth having:

```
AsciiDoc  Elastic  Google  Hugo  Joblint  MDX  Microsoft  NoAnimalViolence
OpenShiftAsciiDoc  Readability  RedHat  Salesforce  alex  neighbor
proselint  write-good
```

Google, Microsoft, Red Hat, Salesforce, and Elastic are done, MIT-licensed, and attributed upstream. Building a second registry against a schema nobody uses yet would be re-implementing that.

## License audit

Read from each project's `LICENSE` file or page footer, not from memory:

| Guide | License | Shippable here |
|---|---|---|
| Google developer documentation | CC BY 4.0 | Yes, with attribution |
| Microsoft Writing Style Guide ([repo](https://github.com/MicrosoftDocs/microsoft-style-guide)) | CC BY 4.0 | Yes, with attribution |
| GOV.UK style guide | Open Government Licence v3.0 | Yes, with attribution |
| 18F Content Guide | CC0 1.0 / US public domain | Yes, no conditions |
| Red Hat supplementary style guide | CC BY-SA 4.0 | ShareAlike conflicts with MIT |
| Canonical documentation style guide | CC BY-SA 4.0 | Same |
| Mailchimp Content Style Guide | **CC BY-NC 4.0** | No: NonCommercial |
| Apple Style Guide | "All rights reserved" | No |
| CMOS, APA, MLA, AP | Paywalled, trademarked names | No |

Two that surprised me. Mailchimp's is NonCommercial, despite its reputation as the open one. And the ShareAlike guides would push copyleft onto any derivative config, which conflicts with this repo's MIT.

[...]

Issue #76, opened 2026-07-31
Proposal: --style layer


ClaySecAI:
**Motivation.** The skill removes AI-isms but can't target a specific style guide. There is no way to make output follow Google, Chicago, or APA conventions.

**Proposal.** An optional `--style none | google | cmos | apa` layer on top of the current pass. `none` is the default and changes nothing. Each value adds a register and the mechanics for one guide. Skill-only, with no new detector types, so the category counts stay at 60/112. The same pattern extends to an organization's own conventions: a team could add its house style as another `--style` value.

**What a small eval showed (directional: n=1 per genre, LLM-judged).** In a genre × style matrix (blind 3-judge panel, rubric matched to each genre), each style scored highest on its home genre and worse off-genre, so the layer has to be opt-in and genre-aware. On clean technical input the plain pass already scored well; the style layer's gain there was mechanical consistency, such as sentence-case headings, not a register fix. Separately, a rewrite emits straight quotes even when a guide wants curly, so guide-consistency needs a deterministic marks pass (a small normalizer, tested).

**Question.** Is a `--style` layer in scope for this skill? I have a working implementation ready (backward-compatible, no changes when `none`, gates green, tests) and will open a PR if it fits.

conorbronsdon:
Great idea @ClaySecAI feel free to submit a PR and I will adapt as needed

conorbronsdon:
Following up on my "send a PR" with something more useful than a blank cheque, since you said the implementation is already working and I'd rather narrow the target before you spend more time on it than after.

**The marks pass I'll take as-is.** You found a real defect: the rewrite emits straight quotes regardless of what the surrounding document uses. A small deterministic normalizer for quotes, dashes, and heading case is mechanical, testable, and doesn't depend on any external guide. Send that on its own and it can land quickly. Sentence-case headings interact with `title-case-header` and with the carve-out in `detector/validate.js` that exempts heading rewrites, so worth a fixture pinning that interaction.

**The three-guide layer is where I want to push back**, and your own eval is most of the reason.

You reported that on clean technical input the plain pass already scored well and the style layer's gain was "mechanical consistency, such as sentence-case headings, not a register fix." That reads to me as the measurable win being the normalizer above, with the register part unmeasured — and the register part is what costs three external guides.

Three specific problems on this end:

[...]

PR #99, opened 2026-08-03
feat: optional --style house-style layer (config-driven, points at Vale)

ClaySecAI:
## Summary

Implements the config-driven `--style` layer from #76, in the shape decision 1 of #88 records. Reviewed on my fork first (ClaySecAI/avoid-ai-writing#3), rebased onto `main` at 3.22.3, version bumped to 3.23.0 across SKILL.md frontmatter, `package.json` and `plugin.json` with both generated copies regenerated.

`--style ./house.json` applies a user-supplied JSON config: `register` directives the model follows, plus `mechanics` that `scripts/check-style.js` checks. Only `quotes` and `latinAbbrev` gate the exit code (0 clean / 1 hard / 2 tool error); `headings`, `emDash` and `spellNumbersUpTo` are advisory and `serialComma` is never checked, so the docs scope the verb to the checkable subset rather than claiming "mechanics are verified".

No bundled guides: `examples/` holds two generic starters and the schema, and links #88; the README section from #89 already says where encoded guides live. A bare `--style "APA"` is a fallback rather than a feature, since SKILL.md instructs the model to claim no compliance and not to reproduce the guide's text, and an instruction is not a checked rule. `--style` is in the invocation line with an explicit composition rule against `--voice` and `--context`.

## Existing files this changes

- **`README.md`** said "There's no `--style` input", which this makes false. The first paragraph now describes the config input and reconciles with your heading, since enforcing a *published* guide is the different job. Your Vale paragraph is untouched, the paywalled one gains two sentences, and the heading is unchanged because `examples/README.md` links its anchor.
- **`scripts/sync-cursor-rules.sh`** gains two anchored spans, the way span 3 handles `validate.js`. Without them the ported rule tells a Cursor user to run `scripts/check-style.js` and read `examples/`, neither of which ships in the `.mdc`. Worth flagging: the drift guard catches edits to anchored spans but not a new section introducing repo references, which is how I nearly shipped a broken port.
- **Wiring**: `detector-test.yml` path filters widen to `scripts/**` and add `examples/**` on both triggers; `package.json` appends `check-style.test.js`.

## False positives

Four adversarial passes, each finding that the previous pass's fix had introduced a new way to hard-fail a correct document or miss a real one. All are now tests: markdown link titles and reference definitions (straight quotes are *syntax* there, and the shipped `prose.json` sets `curly`), HTML attributes without swallowing prose containing a comparison (`n<N ... >`), URLs with parentheses, nested and tilde fences, BOM'd frontmatter, a parenthetical spanning a code block, and a stray `(` that silently suppressed every later `latinAbbrev` finding in the document.

Masking runs refdef, then links, then tags, because masking a destination first makes `[a](url): text` re-parse as a reference definition. Link masking walks to the matching paren and scans linearly, replacing a quadratic regex.

## Two things I'd rather you decide

- **Bare names.** `--style technical` is a checked config and `--style chicago` the fallback, with only which filenames exist separating them. If `google.json` ever ships under decision 2, `--style google` silently becomes verified. Dropping bare names (a path is a config, anything else is the fallback) also removes the shadowing and traversal guards. I left it as reviewed and would happily change it.
- **The plugin bundle** ships SKILL.md alone, so its `--style` section points at files the bundle doesn't carry. That matches precedent, since SKILL.md already references `detector/validate.js`, so I left it rather than invent a third portability rule.

## Verification

```
npm test                 detector, categories, validate, corpus, check-style (32)   exit 0
self-scan:check          SKILL.md 14630w raw 82 exempt 13 budget 25                 PASS
                         README.md  4253w raw 64 exempt 21 budget 30                PASS
check-pattern-count.sh   pattern 61 · word-table 112 (59+40+13)                     in sync
sync-plugin-skill.sh     synced: plugin skill + version (3.23.0)                    exit 0
sync-cursor-rules.sh     synced: cursor rule (v3.23.0); second run byte-identical   exit 0
```

Negative control on the new spans, matching #98: mutating the span-4 text in SKILL.md gives `exit 1`, "anchor for span 4 (check-style.js config path) found 0 times"; restore gives `exit 0`. Leftover repo references in the generated rule: `grep -c 'check-style.js|examples/|validate.js|CATEGORIES.md'` = 0. Added prose scores 0 to 2 under the detector with no new flag types; the em-dash flag on `examples/README.md` is its table and definition-list separators.

## Checklist

- [ ] Added a detector `type`: n/a, no `detector/` file is touched and the catalog stays 61 / 112
- [ ] Added a judgment-only rule: n/a, this is an input axis rather than a detection rule
- [x] I considered false positives and added carve-outs for legitimate human writing (nearly half the 32 tests are must-not-fire cases, with a matching set that must still fire)
- [x] Any factual claim about how AI or humans write cites a source: n/a, this PR makes none, and declines to claim guide compliance it can't measure
- [x] The prose I added passes the skill's own audit
- [x] `CHANGELOG.md` entry added under a dated `## [3.23.0]` heading, and `SKILL.md` `version:` bumped


conorbronsdon:
Pushed the review fixes onto your branch as f900671 rather than round-tripping a change list; reasoning per edit below. Review was two passes: an independent adversarial pass re-ran your whole verification table (it reproduces exactly, including the span-4 negative control and the byte-identical second sync run) and ran a mutation panel against your tests — 6 of 7 mutants killed, which is the best result a test suite here has posted.

What changed and why:

- **Indented code blocks are now masked.** Four-space-indented content after a blank line — pasted terminal output is the everyday case — hard-failed `quotes: curly` and `latinAbbrev` on correct CommonMark, and neither limitations list named it. That was the one real collision with CONTRIBUTING's precision-over-recall bar. Lazy continuation and list-item content stay checked (both pinned with must-fire tests); a code block nested inside a list item needs a fence, and the docs now say so.
- **A leading thematic break is no longer frontmatter.** `---` plus a blank line at the top swallowed everything to the next `---` in the file, silently. Frontmatter now requires a non-blank second line; there's a must-fire pin and a must-still-mask complement.
- **The bare-name traversal regex is deleted, honestly.** Removing it left all 32 of your tests green: `resolveConfig('../package')` contains a separator, so it takes the path branch and never reaches the guard. Containment holds by construction — a bare name is a single path segment — which the comment now states, with a test pinning dot-bearing names. The mutation panel is what surfaced this; your own fix to my `resolveConfig('package')` pin was the same class of problem, so I suspect you'll appreciate the symmetry.
- **Config runs announce themselves**: `Applying config examples/technical.json; checkable mechanics verified.`, mirroring the fallback's status line. This is half of my answer to your first question, below.
- Housekeeping: the double-backtick-span and next-line-refdef-title gaps joined both limitations lists (both narrow; documenting beat complicating the masking), `examples/README.md` joined the self-scan budgets, the CHANGELOG counts updated, and the cursor-rule span anchors follow the SKILL.md edit.

Negative controls on my own additions, same bar you used: the two behavior fixes fail against your pre-fix code (indented-code must-not-fire and leading-HR must-fire both flip FAIL to pass), and the three preservation complements pass on both sides. Full battery after: `npm test` green with check-style at 39, `self-scan:check` PASS, pattern count 61 / 112 in sync, both sync scripts idempotent.

On the two questions you left to me:

1. **Bare names stay.** I went back and forth on your alternative, and what settled it is the failure mode dropping them creates: `--style technical` would route to the named-guide fallback, and the model earnestly applying "technical style" from general knowledge is a worse trap than the shadowing you already fixed. #88's decision 2 also contemplates shipping attributed open-licensed configs later, which makes bare names the eventual primary interface rather than a liability. The upgrade-on-ship hazard you named is real — that's what the mode announcement is for. When a `google.json` someday ships, the output visibly flips from the fallback line to the config line, and the CHANGELOG carries the semantics change.
2. **The plugin bundle stays as-is.** You read the precedent right. Filed #102 for the long-term shape (teach `sync-plugin-skill.sh` the span rewrites, or bundle the dependency-free checker) rather than inventing a third portability rule inside this PR.

Also filed #103 for the drift-guard gap you flagged in the body — a new section introducing repo references is exactly what the anchor checks can't see, and you shouldn't have had to catch it by hand. Your `grep -c = 0` verification is the gate; it just needs to live in the script.

The masking-order comment, the linear paren walk, and the must-fire complements made this fast to review. From my side it's ready to merge once CI confirms.

PR #110, opened 2026-08-04
cursor-rules: fail on repo references a new SKILL.md section introduces (#103)

### LINKED ISSUE #103: Cursor-rule drift guard misses NEW sections that introduce repo references
ClaySecAI flagged this in #99's PR body after nearly shipping it: the cursor-rule drift guard anchors on exact upstream spans, so it catches EDITS to a known span, but a brand-new SKILL.md section that introduces repo references (`scripts/`, `examples/`, `detector/`) sails through untouched into the generated `.mdc`, where those paths mean nothing to a user who curled the rule into their own project.

Cheap gate: after generation, grep the produced `.mdc` for repo-relative references (`detector/`, `scripts/`, `examples/`, `CATEGORIES.md`) and fail loudly on any hit outside the attribution header. #99's own verification already ran exactly that grep by hand (`= 0`); moving it into `sync-cursor-rules.sh` makes it structural, so the next new section can't repeat the near-miss.

ClaySecAI:
## Summary

Closes #103. The anchored rewrites catch EDITS to a span they already know about; a brand-new SKILL.md section that names a repo path passes them untouched and lands in a rule users curl into a project where none of this repo's files exist. That is how the `--style` section nearly shipped pointing at `scripts/check-style.js` and `examples/`, caught by hand in #99 rather than by the guard.

The gate scans the generated body for references into `detector/`, `scripts/`, `examples/`, `plugins/`, `cursor-rules/` and `corpus/`, plus the distinctive bare filenames (`check-style.js`, `validate.js`, `patterns.js`, `self-scan.js`, `CATEGORIES.md`, `PROOF.md`). It exits 1 quoting the match and the line number it would occupy in the rule.

Three details worth the review:

- **Only `body` is scanned.** The attribution header legitimately names `scripts/sync-cursor-rules.sh`, and it lives in `cursor_fm`, so it is exempt by construction. A line-range or regex carve-out would drift the moment that header changes.
- **It runs before the write.** A failing generation leaves no bad artifact on disk to be committed by accident, which is a little stronger than grepping the file afterward.
- **A match has to name a file**, not a bare directory. Precision is the whole difficulty here: the directory names collide with ordinary prose in this document. My first revision matched bare directories and false-positived on unmodified `SKILL.md`, because `docs/technical-blog` is a context pair rather than a path. Requiring the extension is what makes the gate safe to leave in CI, and it still catches every real reference, which always names a file.

Three further exemptions, each a precision call: a left word boundary, so `manuscripts/x.md` stays prose; anything inside an `http(s)` URL, which is the portable form and also what this gate's own error message tells an author to switch to; and `README`/`CHANGELOG`/`CONTRIBUTING`/`CLAUDE`, which every project has and which this skill's prose already names as documents it edits.

## Verification

```
sync-cursor-rules.sh   synced: cursor rule (v3.23.0); no repo references     exit 0
                       second run byte-identical, .mdc unchanged             exit 0
npm test               detector, categories, validate, corpus, check-style   exit 0  (39)
self-scan:check        every document within budget                          PASS
check-pattern-count.sh pattern 61 · word-table 112 (59+40+13)                in sync
sync-plugin-skill.sh   synced: plugin skill + version (3.23.0)               exit 0
```

A 15-case matrix, all correct. Must-pass: `manuscripts/drafts`, `transcripts/notes`, `superscripts/subscripts`, `examples/counterexamples`, `docs/technical-blog`, `README.md`, and two GitHub URLs whose paths contain `detector/` and `examples/`. Must-block: the #99 near-miss shape, `examples/<name>.json`, `detector/CATEGORIES.md`, `./scripts/self-scan.js`, and the bare filenames `validate.js` and `PROOF.md`.

Negative control, the #103 scenario exactly: adding a section containing `node scripts/whatever.js` and `examples/thing.json` to SKILL.md gives

```
sync-cursor-rules: 1 repo-relative reference(s) would ship in the Cursor rule, where
none of this repo's files exist:
  line 721: scripts/whatever.js   in: Run `node scripts/whatever.js` and read `examples/thing...
Add a portability rewrite for the new span (see the header comment), or reword the
SKILL.md section so it stands alone.
```

exit 1, with `cursor-rules/avoid-ai-writing.mdc` unchanged; removing the section gives exit 0. The reported line number was checked against the generated file rather than assumed (body line + header offset resolves to the real `.mdc` line).

No change to the current rule: the only existing match in the whole file is the attribution header.

## Checklist

- [ ] Added a detector `type`: n/a, no `detector/` file is touched and the catalog stays 61 / 112
- [ ] Added a judgment-only rule: n/a, this is build tooling
- [x] I considered false positives and added carve-outs for legitimate human writing: the header exemption is the carve-out, and it is structural rather than pattern-matched
- [x] Any factual claim about how AI or humans write cites a source: n/a, this PR makes none
- [x] The prose I added passes the skill's own audit
- [ ] `CHANGELOG.md` entry and `SKILL.md` `version:` bump: n/a, build tooling with no SKILL.md content change, matching #96 and #98

### DIFF
--- a/scripts/sync-cursor-rules.sh
+++ b/scripts/sync-cursor-rules.sh
@@ -15,12 +15,17 @@
 # Each rewrite is anchored on the exact upstream text and FAILS LOUDLY if the
 # anchor stops matching exactly once — so an upstream edit to one of those
 # spans breaks CI here instead of silently shipping a wrong Cursor rule.
+#
+# The anchors only see spans they already know about. A brand-new SKILL.md
+# section that introduces a repo path passes them untouched, so a final gate
+# greps the generated body for repo-relative references and fails on any hit.
 set -euo pipefail
 
 repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 
 python3 - "$repo_root/SKILL.md" "$repo_root/cursor-rules/avoid-ai-writing.mdc" <<'PY'
 import io
+import re
 import sys
 
 src, dst = sys.argv[1], sys.argv[2]
@@ -98,6 +103,57 @@ alwaysApply: false
      scripts/sync-cursor-rules.sh; CI fails when the two drift. -->
 """
 
+# ── Portability gate (#103) ──────────────────────────────────────────
+# The anchored rewrites above catch EDITS to a span they already know. A new
+# SKILL.md section naming a repo path sails through them and lands in a rule
+# that users curl into a project where this repo does not exist. Only `body`
+# is scanned: the attribution header names scripts/sync-cursor-rules.sh and
+# lives in cursor_fm, so it is exempt by construction rather than by carve-out.
+#
+# A reference worth blocking names a FILE: `scripts/check-style.js`,
+# `examples/<name>.json`, `detector/CATEGORIES.md`. Requiring the extension is
+# what keeps the gate precise, because the bare directory names collide with
+# ordinary prose in this document: "docs/technical-blog" is a context pair,
+# "examples/counterexamples" and "transcripts/notes" are slash-joined words.
+# Two further exemptions, each a precision call:
+#   - a left word boundary, so "manuscripts/x.md" is prose, not scripts/
+#   - anything inside an http(s) URL, which is the portable form and is also
+#     what this gate's own error message tells an author to switch to
+REPO_DIRS = ("detector", "scripts", "examples", "plugins", "cursor-rules", "corpus")
+# Distinctive enough to block unqualified; README.md and friends are omitted on
+# purpose, since every project has them and this skill's prose names them.
+REPO_FILES = ("check-style.js", "validate.js", "patterns.js", "self-scan.js",
+              "CATEGORIES.md", "PROOF.md")
+FILENAME = r"(?:<[^<>\s]+>|[A-Za-z0-9_.-]+)\.[A-Za-z0-9]+"
+REPO_REF = re.compile(
+    r"(?<![A-Za-z0-9])(?:"
+    + r"(?:" + "|".join(re.escape(d) for d in REPO_DIRS) + r")/" + FILENAME
+    + r"|" + "|".join(re.escape(f) for f in REPO_FILES)
+    + r")"
+)
+URL = re.compile(r"https?://\S+")
+
+leaks = []
+for n, line in enumerate(body.split("\n"), 1):
+    m = REPO_REF.search(URL.sub("", line))
+    if m:
+        leaks.append((n, m.group(0), line.strip()))
+if leaks:
+    offset = cursor_fm.count("\n")
+    shown = "\n".join(
+        # Quote the match itself: a long line truncated from the left can hide it.
+        f"  line {offset + n}: {hit}   in: {text[:72]}{'...' if len(text) > 72 else ''}"
+        for n, hit, text in leaks[:10]
+    )
+    more = f"\n  ... and {len(leaks) - 10} more" if len(leaks) > 10 else ""
+    sys.exit(
+        f"sync-cursor-rules: {len(leaks)} repo-relative reference(s) would ship in the "
+        f"Cursor rule, where none of this repo's files exist. Line numbers are for the "
+        f"rule that would have been generated:\n{shown}{more}\n"
+        "Add a portability rewrite for the new span (see the header comment), link the "
+        "full https://github.com/... URL, or reword the section so it stands alone."
+    )
+
 io.open(dst, "w", encoding="utf-8", newline="\n").write(cursor_fm + body)
-print(f"synced: cursor rule (v{version})")
+print(f"synced: cursor rule (v{version}); no repo references in the ported body")
 PY


conorbronsdon:
Reviewed against a fresh checkout of `8d52369` — every claim re-run, plus some probes beyond the PR's matrix. The gate's design is sound and the precision engineering is real, but two findings should be fixed before merge.

## Verified clean (independently re-run)

- Clean tree: `sync-cursor-rules.sh` exit 0, `.mdc` **byte-identical** across runs; the only pre-existing match in the whole file is the attribution header, which is exempt by construction as described.
- Must-fire control (the exact #103 shape, new section with `node scripts/whatever.js` + `examples/thing.json`): exit 1, match quoted with a generated-file line number, **`.mdc` left untouched** — the gate runs before the write, as claimed.
- Precision matrix: all of the PR's must-pass and must-block cases behave as stated when re-tested directly against the regex — `manuscripts/x.md`, `docs/technical-blog`, `examples/counterexamples`, `transcripts/notes`, `superscripts/subscripts`, README/CHANGELOG/CONTRIBUTING, both URL cases pass; `./scripts/self-scan.js`, `detector/CATEGORIES.md`, bare `validate.js`/`PROOF.md`, `corpus/`, `plugins/`, `cursor-rules/` file refs all block. Left word boundary holds (`DISPROOF.md` does not match).
- Line-number arithmetic checks out (1-based body line + `cursor_fm` newline count).
- `npm test` (all five suites), `self-scan:check` PASS, `check-pattern-count.sh` in sync (61 / 112), `sync-plugin-skill.sh` clean tree after run. CI green here too.

## Fix before merge

**1. One verification claim in the PR body is false (confidence 85).** The must-block list includes `examples/.json`, and the body says "A 15-case matrix, all correct." The shipped regex does **not** match `examples/.json`: `FILENAME` requires at least one name character before the extension dot, and backtracking can't rescue a bare `.json` (the character class consumes the only dot, leaving no literal `.` to match). Re-testable in one line:

```python
REPO_REF.search("examples/.json")   # -> None
```

The *behavior* is defensible — a bare dotfile isn't one of this repo's files — but then the matrix mislabels the case. Fix either side: correct the PR body (move it to must-pass with a sentence on why), or require it to block. House rule here is that claim scope = verification scope, so this one is a blocker as written.

**2. The bare-filename alternation has no right boundary (confidence 80).** `patterns.json` in prose matches as `patterns.js`, `validate.json` as `validate.js` — the gate would block a future section that legitimately mentions a user-side `*.json` with one of those stems, and quote a match that isn't what the author wrote. It fails closed (blocks loudly), so it's a false-positive nuisance rather than a hole, but the fix is one token — a right guard after the alternation:

```python
+ r")(?[A-Za-z0-9])"
```

(`patterns.js,` and `patterns.js` at end-of-line still block; `patterns.json` stops matching.) Worth adding `patterns.json`-style cases to the matrix as must-pass when you do.

## Noting, not blocking

- Per-line `search()` reports only the first match on a line, so the count in the error message can undercount (my must-fire line carried two refs; the message said "1 reference(s)"). The gate still fails, so this is cosmetic — `finditer` would make the count honest if you're touching the file anyway.

Verdict: **NEEDS FIXES** — both are small, and #1 may be a one-paragraph PR-body edit. Happy to re-verify on a new push.

---

conorbronsdon:
Thanks for this — the premise checks out (I reproduced #103 against main: a new section with repo refs sails into the generated `.mdc` with exit 0), and the fail-closed shape is right: a failing run leaves the on-disk artifact untouched, and your clean-path claim is true — regenerated `.mdc` came out byte-identical to the committed one under the PR's script.

One change needed before merge: **`FILENAME` can't cross a `/`, so nested paths escape the gate.** Both of these pass exit 0 today:

```
plugins/avoid-ai-writing/SKILL.md
corpus/human/essay-01.md
```

Every real file under `plugins/` in this repo is nested (there are zero flat files there), so the `plugins` entry in `REPO_DIRS` currently reads as a guarantee that can never fire. Allowing path segments before the final filename (and adding a nested must-fire case next to your existing matrix) closes it.

Three smaller things I hit while testing — fine to fold in or leave for a follow-up, all latent (none occur in the current SKILL.md, and all fail closed):

1. Bare filenames have no right boundary: `patterns.json` trips via its `patterns.js` prefix.
2. The left lookbehind omits `_`, so `my_scripts/foo.js` fires.
3. Markdown link text fires even when the link target is the full https URL — and the error message recommends exactly that form, so the message sends authors somewhere the gate still blocks. Worth aligning one or the other.

The 15-case matrix in the PR body was genuinely useful for review — with the nested fix it'd be worth committing a small fixture test so it keeps running.

conorbronsdon:
Merged this along with a batch today and only afterwards caught that the two review comments above were still unaddressed -- that's on me, not on @ClaySecAI. The gate as merged is sound on the flat cases (I'd independently re-verified fire/no-fire before merging); the outstanding findings (nested paths, boundaries, link-form alignment) plus the committed self-test matrix are in #114. Thanks again for the contribution -- the precision engineering held up under two adversarial passes, which is rare.

PR #106, opened 2026-08-04
docs: refresh repository guidance

JasonColapietro:
## Summary

- refresh `CLAUDE.md` to describe the current detector and test workflow
- remove the duplicated skill version so it cannot drift again
- update the pattern-category guidance and guard that copy through `.ssot.yaml`

## Testing

- `npm test`
- `git diff --check`

Closes #95.

Issue #95, opened 2026-08-03
CLAUDE.md is stale on three facts, including the pattern count in the sentence telling you not to restate it


conorbronsdon:
`CLAUDE.md` is the first file an agent reads in this repo, and it is stale on three independent facts. Two of them are counts, which is the same class #91 just fixed everywhere else.

| line | says | actual |
|---|---|---|
| `CLAUDE.md:11` | ``SKILL.md`` is `v3.7.2` | `3.22.2` |
| `CLAUDE.md:23` | the README bullet reads "46 pattern categories" | it reads **61** |
| `CLAUDE.md:7` | "No build system, no dependencies, no tests" | `npm test` runs four suites |

The `:23` one is the pointed one. That line exists to say:

> The pattern count lives in **one** place — the README "46 pattern categories" bullet — and is derived from SKILL.md's detection `###` entries. Don't restate it elsewhere.

It restates the count in the sentence telling you not to restate the count, and gets it wrong by 15. That is not a typo, it is the failure mode the file is warning about, demonstrated in place.

Reproduce:

```
$ grep -n "^version:" SKILL.md
4:version: 3.22.2
$ grep -n "pattern categories\*\*" README.md
43:- **61 pattern categories** — ...
$ npm test
# detector fixtures, CATEGORIES.md contract, preservation validator, corpus
```

## Why it was not caught

`.ssot.yaml` tracks the catalog count with `README.md:43` as canonical and lists the promo surfaces as copies. `CLAUDE.md` is not one of them, so `ssot_check.py check` cannot see it. The `#91` guard added to `detector/categories.test.js` covers the *engine type* count and not this one.

## Proposed fix

Correct all three, then decide which deserve a guard rather than an edit:

- **The version.** `scripts/sync-plugin-skill.sh` already propagates `SKILL.md` frontmatter to `plugin.json`. Either add `CLAUDE.md` to what it syncs, or delete the version from `CLAUDE.md` — it does no explanatory work there and the frontmatter is one file away.
- **The catalog count.** Add `CLAUDE.md` as a copy in `.ssot.yaml` under `avoid-ai-pattern-count`. That is exactly what the manifest is for, and unlike the engine count this one *is* a copied fact with a prose canonical. Best option of the three.
- **"No tests."** Just wrong, and nothing can guard a sentence like that. Reword to name the suites.

I would take the `.ssot.yaml` route for the count and delete the version outright.

PR #111, opened 2026-08-04
check-pattern-count: assert CLAUDE.md's category-count copy on every push

conorbronsdon:
## Summary

Closes the latency gap noted while reviewing #106: CLAUDE.md's pattern-count copy is tracked by `.ssot.yaml`, but that manifest only runs at release / weekly cron (promo-drift), so a same-repo edit can sit stale for up to a week — the exact rot #95 documented. `check-pattern-count.sh` now asserts the copy in CI, and after the review-round fix the workflow actually triggers on the edits the guard exists to catch: `CLAUDE.md` joined `plugin-skill-sync.yml`'s `paths:` filter on both the PR trigger and the push-to-main trigger (previously a CLAUDE.md-only change ran zero checks — verified on PR #106).

Extraction is `grep -o | head -n1` (first match in file order — the same match a first-match regex engine reading `.ssot.yaml`'s pattern takes; the first draft's greedy sed would have taken the *last* match on a line). A missing CLAUDE.md gets its own message and exit 1 instead of dying on sed's raw error, since the file is a tracked `.ssot.yaml` copy and a fork that deletes it should update both together.

## Testing (exit codes checked unpiped)

```
clean tree                        pattern 61 · CLAUDE.md copy 61 · word-table 112   exit 0
CLAUDE.md 61 -> 62                "CLAUDE.md quotes 62, SKILL.md derives 61"        exit 1
sentence reworded                 "could not find ... update .ssot.yaml together"   exit 1
CLAUDE.md deleted                 "CLAUDE.md not found — tracked copy (.ssot.yaml)" exit 1
same line mentions 61 then 46     first match wins: copy reads 61                   exit 0
```

`npm test` unaffected (no detector or SKILL.md change; catalog stays 61 / 112). No CHANGELOG entry or version bump — build tooling only, matching #96/#98/#110 precedent.

Review history: the adversarial pass on the first draft found the workflow-trigger gap (HIGH), the dead missing-file branch (MEDIUM), and the greedy-sed last-match semantics (LOW) — all fixed in `237a089`.

https://claude.ai/code/session_01EKwz2XfwaMD6uR5ABmVaKq

### DIFF
--- a/.github/workflows/plugin-skill-sync.yml
+++ b/.github/workflows/plugin-skill-sync.yml
@@ -5,6 +5,10 @@ on:
     paths:
       - "SKILL.md"
       - "README.md"
+      # CLAUDE.md carries a checked copy of the pattern count; without this
+      # path the count guard never runs on a CLAUDE.md-only edit (PR #106
+      # had zero check runs - review finding on #111).
+      - "CLAUDE.md"
       - "plugins/**"
       - ".claude-plugin/**"
       - "cursor-rules/**"
@@ -14,6 +18,10 @@ on:
     paths:
       - "SKILL.md"
       - "README.md"
+      # CLAUDE.md carries a checked copy of the pattern count; without this
+      # path the count guard never runs on a CLAUDE.md-only edit (PR #106
+      # had zero check runs - review finding on #111).
+      - "CLAUDE.md"
       - "plugins/**"
       - ".claude-plugin/**"
       - "cursor-rules/**"
--- a/scripts/check-pattern-count.sh
+++ b/scripts/check-pattern-count.sh
@@ -47,11 +47,19 @@ echo "pattern count in sync: $detection_count"
 # The count has one same-repo copy: CLAUDE.md's "don't restate it" sentence
 # quotes the README bullet. .ssot.yaml tracks that copy, but the manifest only
 # runs on release / weekly cron (promo-drift), so a same-repo edit could sit
-# stale for up to a week. Asserting it here closes the window on every push.
-# The sed pattern mirrors .ssot.yaml's regex for the copy, so the two guards
-# cannot read different sentences.
+# stale for up to a week. Asserting it in CI closes the window. Extraction is
+# grep -o + head, which takes the FIRST match in file order — the same match
+# a standard first-match regex engine reading .ssot.yaml's pattern would take
+# (a greedy sed s/.*…/ would silently take the LAST match on a line instead).
 claudemd="$repo_root/CLAUDE.md"
-claude_count="$(sed -n 's/.*README "\([0-9][0-9]*\) pattern categories" bullet.*/\1/p' "$claudemd" | head -n1)"
+
+if [ ! -f "$claudemd" ]; then
+  echo "CLAUDE.md not found — it carries a tracked copy of the pattern count (.ssot.yaml)." >&2
+  echo "If it was deliberately removed, drop the copy from .ssot.yaml and this check together." >&2
+  exit 1
+fi
+
+claude_count="$(grep -o 'README "[0-9][0-9]* pattern categories" bullet' "$claudemd" | head -n1 | tr -cd '0-9' || true)"
 
 if [ -z "$claude_count" ]; then
   echo "could not find the 'README \"NN pattern categories\" bullet' sentence in CLAUDE.md" >&2
--- a/scripts/check-pattern-count.sh
+++ b/scripts/check-pattern-count.sh
@@ -44,6 +44,29 @@ fi
 
 echo "pattern count in sync: $detection_count"
 
+# The count has one same-repo copy: CLAUDE.md's "don't restate it" sentence
+# quotes the README bullet. .ssot.yaml tracks that copy, but the manifest only
+# runs on release / weekly cron (promo-drift), so a same-repo edit could sit
+# stale for up to a week. Asserting it here closes the window on every push.
+# The sed pattern mirrors .ssot.yaml's regex for the copy, so the two guards
+# cannot read different sentences.
+claudemd="$repo_root/CLAUDE.md"
+claude_count="$(sed -n 's/.*README "\([0-9][0-9]*\) pattern categories" bullet.*/\1/p' "$claudemd" | head -n1)"
+
+if [ -z "$claude_count" ]; then
+  echo "could not find the 'README \"NN pattern categories\" bullet' sentence in CLAUDE.md" >&2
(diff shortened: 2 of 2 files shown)


conorbronsdon:
## Adversarial review

Ran every claim in the PR body against the branch at `e9b0cc1`. The assertion itself works exactly as described. The problem is that CI will almost never reach it.

### Verified clean

All three testing-table rows reproduce exactly:

```
$ bash scripts/check-pattern-count.sh                 # clean tree
pattern count in sync: 61
CLAUDE.md copy in sync: 61
word-table count in sync: 112 (59 + 40 + 13)
exit=0

$ sed -i 's/README "61 pattern/README "62 pattern/' CLAUDE.md
pattern-count drift: CLAUDE.md quotes 62, SKILL.md derives 61
exit=1

$ # sentence reworded to 'the README bullet that says 61 pattern categories'
could not find the 'README "NN pattern categories" bullet' sentence in CLAUDE.md
If the sentence was reworded, update this check AND the CLAUDE.md copy in .ssot.yaml together.
exit=1
```

`npm test` exits 0 (`All detector fixtures passed` / `all preservation tests passed` / `all corpus tests passed` / `check-style: 39 passed`). Both CI checks on the PR are green. Quote style matches: CLAUDE.md:24 uses ASCII `"` (verified `0x22` by `od -c`), the same literal `.ssot.yaml`'s `'README "(\d+) pattern categories" bullet'` expects, and there is exactly one match in the file today. CLAUDE.md was restored after each mutation; tree is clean.

### Defect 1 — the guard does not run on a CLAUDE.md-only edit (high)

`plugin-skill-sync.yml` is the only workflow that invokes this script, and its `paths:` filter is:

```yaml
paths: ["SKILL.md", "README.md", "plugins/**", ".claude-plugin/**", "cursor-rules/**", "scripts/**"]
```

`CLAUDE.md` is not in it. Nothing else in the repo calls the script — `grep -rn check-pattern-count . --exclude-dir=.git` returns only this workflow, a comment in `promo-drift.yml`, a comment in `detector/categories.test.js`, and prose in CLAUDE.md and `.ssot.yaml`. There is no pre-commit hook installed (`.githooks/` absent, `.git/hooks/` has only samples).

So the scenario the PR exists to close — "a same-repo edit can sit stale for up to a week" — is exactly the scenario that does not trigger the check. Concretely, the parent commit is #106, which changed `CLAUDE.md` and `.ssot.yaml` and nothing else:

```
$ git show --stat cf60f76
 .ssot.yaml | 2 ++
 CLAUDE.md  | 9 +++++----
```

`get_check_runs` on PR #106 returns `total_count: 0`. Zero checks ran. Had that PR mistyped the count, this new assertion would not have caught it, and the week-long window stays open. Adding `"CLAUDE.md"` to both the `pull_request` and `push` path lists is a two-line fix and is what makes the rest of the PR worth merging.

Related wording: the summary and `.ssot.yaml:5` both say this runs "every push," but the `push` trigger is `branches: [main]` only. Feature-branch pushes rely on the `pull_request` trigger, so the check is per-PR, not per-push.

### Defect 2 — a missing CLAUDE.md dies on the raw sed error, not the friendly one (medium)

The `-z "$claude_count"` branch is unreachable when the file is absent. Under `set -euo pipefail`, the exit status of `claude_count="$(sed ... )"` is the command substitution's status, so the script dies at the assignment:

```
$ mv CLAUDE.md /tmp/ && bash scripts/check-pattern-count.sh
pattern count in sync: 61
sed: can't read /workspace/avoid-ai-writing/CLAUDE.md: No such file or directory
exit=2
```

Exit 2, not 1, and none of the guidance about updating `.ssot.yaml` together is printed. The `-z` message only fires when the file exists but the sentence is gone — that path works, as the reword test above confirms.

On the fork question: hard-failing is the **right** behavior for this repo, so this is a message defect rather than a policy one. CLAUDE.md is tracked here, `.ssot.yaml` already declares it a copy, and a fork that deletes CLAUDE.md while keeping `.ssot.yaml` is already broken for `promo-drift`. Worth noting the README's install path is `git clone` of the whole repo (not just SKILL.md), so forks do get `scripts/` — but they only run it if they opt into Actions. Suggest an explicit `[ -f "$claudemd" ] ||` check so the failure says which of the two problems it is.

### Defect 3 — the two guards *can* read different sentences (low, but the claim is stated absolutely)

The code comment and summary both assert the sed pattern "mirrors `.ssot.yaml`'s regex, so the two guards cannot read different sentences." `head -n1` handles multiple matching *lines*, but the leading `.*` in the sed expression is greedy, so within a single line sed captures the **last** match while a `re.search`-based reader captures the **first**:

```
line: The README "61 pattern categories" bullet is canonical; the old README "46 pattern categories" bullet is retired.

sed -n 's/.*README "\([0-9][0-9]*\) pattern categories" bullet.*/\1/p' | head -n1   ->  46
re.search(r'README "(\d+) pattern categories" bullet')                             ->  61
```

Caveat on scope: I could not read `conorbronsdon/ssot-check` (not authorized for this session), so the `re.search` half is the standard-implementation assumption, not a verified fact about your checker. Independent of that, the greedy-`.*` behavior is worth fixing on its own terms — the `head -n1` shows the intent was first-match-wins, and sed does not deliver that within a line. `sed -n 's/^.\{-\}README...` is not portable; the simplest fix is `grep -o` before the capture, or note the single-occurrence assumption explicitly.

### Verdict: needs changes

Defect 1 is the blocker — as written the check is real but effectively unreachable for the change class it was built to catch, so the PR does not deliver the latency fix it claims. Defects 2 and 3 are small and can ride along. The assertion logic itself is correct and well-tested, and the decision not to bump the CHANGELOG matches the #96/#98/#110 precedent.


---

conorbronsdon:
All three findings confirmed and fixed in `237a089`; PR body rewritten to match reality.

- **HIGH (guard never triggers):** dead right, and the PR #106 zero-check-runs evidence made it undeniable. `CLAUDE.md` added to `plugin-skill-sync.yml`'s `paths:` on both triggers, with a comment citing this finding. The body's "every push" phrasing corrected too.
- **MEDIUM (dead `-z` branch):** explicit `[ -f ]` check with its own message, exit 1. Took your judgment on the fork question as-is — hard-fail is correct here because `.ssot.yaml` already declares the copy; the message now says to update both together on deliberate removal.
- **LOW (greedy sed = last match):** extraction switched to `grep -o | head -n1` (file-order first match), and the "cannot read different sentences" absolutism replaced with an honest statement of the semantics. Added a same-line 61-then-46 control that resolves to 61.

All five controls re-run with exit codes checked unpiped: 0 / 1 / 1 / 1 / 0.

---

PR #113, opened 2026-08-04
release: npm publish with provenance (closes the 3.22.1 freeze gap)

conorbronsdon:
npm sat at 3.22.1 while the repo walked to 3.23.0 because the Release workflow had no publish step at all. This adds a `publish-npm` job after the GitHub-release job: version-match guard (package.json must equal the CHANGELOG heading, else fail loudly — the drift the workflow already warns about must block a publish), npm-idempotency check, `npm test`, then `npm publish --provenance --access public` via Trusted Publishing (OIDC, `id-token: write`).

**One thing to confirm before merge (npm web UI, can't be checked from here):** the package's Trusted Publisher setting must name `release.yml` in this repo — if it names nothing or a different workflow, the publish will 404-on-PUT (the documented trust-config-mismatch signature). If 3.22.1 was published locally, the trusted-publisher config may not exist yet.

**On first merge:** the job will immediately want to publish 3.23.0 (it's on the CHANGELOG head and absent from npm) — which is the point, but it means merging this IS the 3.23.0 npm release. If you'd rather publish 3.23.0 by hand first, the idempotency check will then no-op.

Not run in CI from this branch (workflow changes only take effect on main for push triggers); YAML validated, and every guard is fail-closed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,3 +73,54 @@ jobs:
             --target "$GITHUB_SHA" \
             --title "v$VERSION" \
             --notes-file /tmp/release-notes.md
+
+  # npm stayed frozen at 3.22.1 while the repo walked to 3.23.0 because this
+  # workflow only ever created GitHub releases — there was no publish step at
+  # all. This job closes that gap. It is independently idempotent (skips when
+  # the version is already on npm), so re-runs and already-tagged pushes are
+  # safe, and it publishes package.json's version only after proving it matches
+  # the CHANGELOG heading the release job used — the drift this workflow
+  # already warns about must block a publish, not ride into the registry.
+  publish-npm:
+    needs: release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write # npm Trusted Publishing (OIDC) + --provenance
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          registry-url: https://registry.npmjs.org
+
+      - name: Refuse to publish a version that disagrees with CHANGELOG.md
+        run: |
+          PKG_VERSION="$(node -p "require('./package.json').version")"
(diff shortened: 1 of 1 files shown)


conorbronsdon:
Both Qodo findings are handled in 29c4a59 (verified by independent adversarial passes before integrating):

1. **Changelog-parse guard** — Qodo's mechanism was wrong: the step runs under the default shell (bash -e, no pipefail), so a missing heading left the variable empty and the mismatch error did fire. What was real: that error read 'CHANGELOG.md ()' (misleading), and the step silently depended on pipefail staying off. The parse now uses the same guarded style as the release job, with a specific missing-heading error. Reproduced before/after in both shell modes.
2. **Publish race** — confirmed, and slightly worse than reported: the release job has the same check-then-act shape, workflow_dispatch is an overlap trigger Qodo missed, and out-of-order runs could drag npm's latest tag backwards. Fixed with workflow-level concurrency (group: release, never cancel); the existing exists-guards make a queued duplicate run a green no-op. YAML parse + npm test verified.

conorbronsdon:
Third Qodo finding (the inline 'publishes wrong commit' comment) handled in 935b59d — confirmed, and it's this PR's own recovery scenario: main currently sits 8 commits past the v3.23.0 tag, one of them changing detector/patterns.js, which is in the published file set. A dispatch from main today would ship post-tag code to npm as 3.23.0 while the GitHub release names f666087 — and npm builds provenance from the run's github.sha, so the mismatch would get signed in. It also doesn't need a dispatch: a CHANGELOG typo fix while a version is tagged-but-unpublished reaches the same state via the push trigger.

Fix fails closed (HEAD must be the tag's commit; distinct error for a missing tag), deliberately NOT Qodo's checkout-the-tag alternative — that would fix the tarball while leaving the attestation naming the wrong commit. Guard is gated on the same exists-check as the publish step, so CHANGELOG edits that don't bump the version stay green. Verified against fixtures incl. annotated-tag peeling, shallow-checkout tag fetch, and a green-to-red mutation control.

PR #17, opened 2026-05-28
Distribute as a Claude Code / Cowork plugin

conorbronsdon:
## Summary

Packages the skill as a **single-plugin marketplace** so it can be installed via `/plugin` instead of a bare file clone. This is the only install path that works in **Claude Cowork**, which loads skills exclusively from installed plugins — closing the gap reported in #12.

```bash
/plugin marketplace add conorbronsdon/avoid-ai-writing
/plugin install avoid-ai-writing@conorbronsdon-skills
```

In Cowork: **Customize → Plugins → Add marketplace from GitHub** → install `avoid-ai-writing`. The skill auto-triggers from phrases like "remove AI-isms," same as in Claude Code.

## What's in here

- `.claude-plugin/marketplace.json` — single-plugin marketplace (`conorbronsdon-skills`).
- `plugins/avoid-ai-writing/` — the plugin: `.claude-plugin/plugin.json` + `skills/avoid-ai-writing/SKILL.md`.
- `scripts/sync-plugin-skill.sh` — regenerates the bundled skill from the root `SKILL.md` and enforces a matching `version`.
- `.github/workflows/plugin-skill-sync.yml` — validates manifest JSON and fails on skill drift or version mismatch.
- `.gitattributes` — pins `*.sh` to LF so the CI script can't be broken by CRLF.
- README — replaces the "wrap it yourself" Cowork note with the real plugin install.

## Single source of truth

Root `SKILL.md` stays canonical. The bundled copy at `plugins/avoid-ai-writing/skills/avoid-ai-writing/SKILL.md` is generated — edit the root, run `bash scripts/sync-plugin-skill.sh`, commit. (A copy is unavoidable: clone-to-skills-dir needs `SKILL.md` at the repo root, while plugin discovery needs `skills/<name>/SKILL.md`. The CI guard keeps them identical.)

## Verification

- `claude plugin validate .` and `claude plugin validate ./plugins/avoid-ai-writing` → both pass.
- Full local cycle: `marketplace add` → `install` → `plugin details` (reports `Skills (1): avoid-ai-writing`, v3.4.0) → `uninstall` → `marketplace remove`.
- Negative-tested the guard: a `9.9.9` plugin.json vs `3.4.0` SKILL.md fails the version check.

## Test plan

- [ ] Confirm CI `plugin-skill-sync` passes on this PR.
- [ ] (Optional) Add the marketplace in a real Cowork desktop session and confirm auto-trigger.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #118, opened 2026-08-06
Write the openclaw emoji literally so org plugin sync accepts the skill

kkwanda:
## Summary

The plugin can't be distributed through Claude **Organization settings > Plugins** on Team and Enterprise plans. Org sync skips it with:

> Skill 'skills/avoid-ai-writing' contains a hex/unicode escape or an escaped line break in its SKILL.md frontmatter, which can spell keys some parsers see and others don't. Write the text literally.

The cause is one line of frontmatter:

```yaml
metadata:
  openclaw:
    emoji: "✍️"
```

Org sync rejects escapes in frontmatter because an escaped *key* can parse differently across parsers — one parser sees a key another doesn't. Reasonable rule, and the escape here is incidental rather than intentional.

This PR writes the character instead:

```yaml
    emoji: "✍️"
```

Same emoji, identical once parsed. Only the source representation changes.

## Scope

Two files, one line each. `SKILL.md` is the canonical edit; the plugin copy was regenerated with `scripts/sync-plugin-skill.sh` rather than edited by hand.

The escape has been present since #17 first packaged the project as a plugin, so every released version is affected — pinning to an earlier tag isn't a workaround for anyone hitting this. Only `metadata.openclaw.emoji` is affected; no other escape appears in either frontmatter block.

## Deliberately not included: the version bump and CHANGELOG entry

The checklist asks for both. I've left them out on purpose: merging a version bump to `main` triggers your automated release and npm publish, and `package.json` and `CHANGELOG.md` have to agree on the version for that pipeline to pass. Which version this rides in, and when, is your call — happy to add the entry if you'd rather it came in with the PR.

Nothing else here changes behaviour: no rule added or altered, no detector `type` touched.

## Checklist

- [x] `npm test` passes (39 check-style + engine fixtures + `CATEGORIES.md` contract check), Node 24.13.0
- [ ] N/A — no detector `type` added
- [ ] N/A — no judgment-only rule added
- [x] No false-positive surface: packaging metadata only
- [x] No factual claim about how AI or humans write
- [x] No prose added
- [ ] Deferred to the maintainer — see above

## Context

Found while distributing this skill to a team through org plugin sync, pointing a private marketplace at `plugins/avoid-ai-writing` via a `git-subdir` source. We're running a patched mirror in the meantime and would much rather go back to tracking upstream.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -9,7 +9,7 @@ metadata:
   tags: writing editing voice quality
   agentskills_spec: "1.0"
   openclaw:
-    emoji: "\u270D\uFE0F"
+    emoji: "✍️"
 ---
 
 # Avoid AI Writing — Audit & Rewrite
--- a/plugins/avoid-ai-writing/skills/avoid-ai-writing/SKILL.md
+++ b/plugins/avoid-ai-writing/skills/avoid-ai-writing/SKILL.md
@@ -9,7 +9,7 @@ metadata:
   tags: writing editing voice quality
   agentskills_spec: "1.0"
   openclaw:
-    emoji: "\u270D\uFE0F"
+    emoji: "✍️"
 ---
 
 # Avoid AI Writing — Audit & Rewrite


conorbronsdon (approved):
Thank you @kkwanda! 

Very glad to hear you've been finding value from the skill: out of curiosity, are you able to share the size of the org you're working with? Always love to hear where the skill is providing value.

PR #120, opened 2026-08-07
feat: detect unnecessary hyphenation

### LINKED ISSUE #107: Detect unnecessary hyphenation
I find Claude _loves_ to hyphenate things that really shouldn't be. Not sure if this is something that could be detected by the Skill? I appreciate this is probably quite tricky and context dependent, but thought I would raise it anyway as it is one of the main things this Skill currently can't fix for me.

Absolutely love this Skill. Use it daily. Thank you 🙏

mbhall88:
## Summary

- add a P2 `unnecessary-hyphenation` detector for the bounded subclasses agreed in #107: welded open noun phrases, curated closed compounds, and position-sensitive compounds
- preserve established compound modifiers and mask code, quotes, blockquotes, URLs, paths, filenames, and command flags before scanning
- split the existing modifier-stacking guidance from grammatical cleanup, bump the catalog to 62 categories and the engine to 48 types, and synchronize the plugin and Cursor copies

The optional `-ly` adverb cleanup is intentionally left out. The issue discussion allows that to arrive later as opt-out style cleanup rather than as an AI tell.

Closes #107.

## User impact

Detect and rewrite suggestions now cover forms such as:

- `research-impact aggregator` → `research impact aggregator`
- `code-base` → `codebase`
- `in real-time` → `in real time`
- `works out-of-the-box` → `works out of the box`

Legitimate uses such as `real-time analytics`, `long-term plan`, `out-of-the-box support`, `high-quality`, and `family-owned` remain unchanged.

## Validation

- `npm test`
- `npm run self-scan:check`
- `bash scripts/check-pattern-count.sh`
- generated plugin and Cursor synchronization checks
- `git diff --check`

## Checklist

- [x] `npm test` passes
- [x] New detector type is documented in `detector/CATEGORIES.md`
- [x] True-positive and must-not-fire fixtures are included
- [x] False positives and protected spans have explicit carve-outs
- [x] `CHANGELOG.md` and version metadata are updated
- [x] Added prose passes the repository self-scan budget

### DIFF
--- a/detector/patterns.js
+++ b/detector/patterns.js
@@ -342,10 +342,10 @@ const AIDetector = (() => {
     'ai-placeholder': 10,
     'ai-citation-markup': 15,
     'ai-utm-source': 12,
-    // Curated copyedits, not authorship evidence. These contribute the same
-    // modest weight as other P2 language cleanups and never act as a trinary
-    // classifier corroborator.
-    'unnecessary-hyphenation': 2,
+    // Curated P2 copyedits, not authorship evidence. Keep them visible in the
+    // issue list without moving the AI score, label, probabilities, or
+    // trinary classification on short documents.
+    'unnecessary-hyphenation': 0,
   };
 
   // ─── Transition phrases ────────────────────────────────────────────
@@ -749,12 +749,107 @@ const AIDetector = (() => {
     }
   }
 
+  function maskYamlFrontmatter(chars) {
+    const lines = chars.join('').split('\n');
+    const bare = (line) => line.replace(/\r$/, '');
+    const first = bare(lines[0]).replace(/^\uFEFF/, '');
+    if (first !== '---' || lines.length < 2 || /^\s*$/.test(bare(lines[1]))) return;
+
+    let closingLine = -1;
+    for (let i = 1; i < lines.length; i += 1) {
+      if (bare(lines[i]) === '---') {
+        closingLine = i;
+        break;
+      }
+    }
+    if (closingLine === -1) return;
+
+    let end = 0;
+    for (let i = 0; i <= closingLine; i += 1) {
+      end += lines[i].length;
+      if (i < lines.length - 1) end += 1;
+    }
+    blankRange(chars, 0, end);
+  }
+
+  function maskYamlMetadata(chars) {
+    const lines = chars.join('').split('\n');
+    let offset = 0;
+    let nestedAfterIndent = null;
+    for (const line of lines) {
+      const bare = line.replace(/\r$/, '');
+      // Lowercase keys are the common unfenced-YAML shape. Keeping this
+      // case-sensitive prevents prose labels such as "Note: ..." from being
+      // mistaken for metadata and silencing a real copyedit on the line.
+      const key = bare.match(/^([ \t]*)(?:-[ \t]+)?[a-z_][a-z0-9_.-]*[ \t]*:(.*)$/);
+      const indentation = (bare.match(/^[ \t]*/) || [''])[0].length;
+      let shouldMask = false;
+
+      if (key) {
+        shouldMask = true;
+        nestedAfterIndent = key[2].trim() === '' ? key[1].length : null;
+      } else if (nestedAfterIndent !== null && bare.trim() !== '' && indentation > nestedAfterIndent) {
+        shouldMask = true;
+      } else if (bare.trim() === '') {
+        nestedAfterIndent = null;
+      } else {
+        nestedAfterIndent = null;
+      }
+
+      if (shouldMask) blankRange(chars, offset, offset + line.length);
+      offset += line.length + 1;
+    }
+  }
+
+  function maskMarkdownTables(chars) {
+    const lines = chars.join('').split('\n');
+    const delimiter = /^\s*\|?\s*:?-{3,}:?\s*(?:\|\s*:?-{3,}:?\s*)+\|?\s*\r?$/;
+    const rows = new Set();
+    for (let i = 0; i < lines.length; i += 1) {
+      if (!delimiter.test(lines[i])) continue;
+      if (i > 0 && lines[i - 1].includes('|')) rows.add(i - 1);
+      rows.add(i);
+      for (let j = i + 1; j < lines.length && lines[j].includes('|'); j += 1) rows.add(j);
+    }
+
+    let offset = 0;
+    for (let i = 0; i < lines.length; i += 1) {
+      if (rows.has(i)) blankRange(chars, offset, offset + lines[i].length);
+      offset += lines[i].length + 1;
+    }
+  }
+
+  function maskDelimitedQuotes(chars, open, close, apostropheAware = false) {
+    const isWord = (char) => char !== undefined && /[a-z0-9]/i.test(char);
+    const isEscaped = (index) => {
+      let slashes = 0;
+      for (let i = index - 1; i >= 0 && chars[i] === '\\'; i -= 1) slashes += 1;
+      return slashes % 2 === 1;
+    };
+    let start = -1;
+    for (let i = 0; i < chars.length; i += 1) {
+      if (start === -1) {
+        if (chars[i] === open && !isEscaped(i) && (!apostropheAware || !isWord(chars[i - 1]))) start = i;
+      } else if (chars[i] === close && !isEscaped(i) && (!apostropheAware || !isWord(chars[i + 1]))) {
+        blankRange(chars, start, i + 1);
+        start = -1;
+      }
+    }
+  }
+
   // Additional protected spans for the hyphenation copyedit. Unlike general
   // AI-tell matching, this rule must not "correct" a literal spelling in a
   // quote, URL, path, filename, command flag, or Markdown blockquote.
   function maskHyphenationProtected(text) {
     const chars = maskCode(text).split('');
     maskTopLevelIndentedCode(chars);
+    maskYamlFrontmatter(chars);
+    maskYamlMetadata(chars);
+    maskMarkdownTables(chars);
+    maskDelimitedQuotes(chars, '"', '"');
+    maskDelimitedQuotes(chars, '“', '”');
+    maskDelimitedQuotes(chars, "'", "'", true);
+    maskDelimitedQuotes(chars, '‘', '’', true);
     const maskMatches = (regex) => {
       const source = chars.join('');
       let match;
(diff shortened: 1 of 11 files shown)


detector/patterns.js
```diff
@@ -709,23 +719,81 @@ const AIDetector = (() => {
...
+        blankRange(chars, match.index, match.index + match[0].length);
+      }
+    };
+
+    maskMatches(/^[ \t]*>[^\n]*$/gm);
+    maskMatches(/"[^"\n]+"|“[^”\n]+”|(?<![a-z0-9])'(?:[^'\n]|(?<=[a-z0-9])'(?=[a-z0-9]))+'(?![a-z0-9])|‘(?:[^’\n]|(?<=[a-z0-9])’(?=[a-z0-9]))+’/gi);
+    maskMatches(/\b(?:https?:\/\/|www\.)[^\s<>()]+/gi);
+    maskMatches(/(?<!\S)--?[a-z0-9][a-z0-9-]*/gi);
+    maskMatches(/(?:[a-z]:[\\/]|(?:\.\.?[\\/])?)(?:[a-z0-9_.-]+[\\/])+[a-z0-9_.-]+/gi);
+    maskMatches(/\b[a-z0-9_.-]+-[a-z0-9_.-]+\.[a-z0-9]{1,16}\b/gi);
```
conorbronsdon: Bounded quantifiers on both masks. Measured linear (~2x per doubling), worst case 5,470 ms → 52 ms through `analyzeText`, zero masking behavior changes across the probe set (`my-file.md`, `foo-bar.test.js`, `some_file-v2.tar.gz` still masked; the attack blob no longer degenerates):

```suggestion
    maskMatches(/(?:[a-z]:[\\/]|(?:\.\.?[\\/])?)(?:[a-z0-9_.-]{1,64}[\\/])+[a-z0-9_.-]{1,64}/gi);
    maskMatches(/\b[a-z0-9_.-]{1,64}-[a-z0-9_.-]{1,64}\.[a-z0-9]{1,16}\b/gi);
```

Please pair this with a timing-asserted regression test on the worst-case input — a correctness-only fixture passes either way.

conorbronsdon (changes requested):
Michael — this is careful work, and review confirmed it: the sync copies are byte-identical, the counts are arithmetically right, the rename left nothing stale, and the scope matches what we settled in #107, including leaving `-ly` out. Because this detector runs on user-supplied text in production, I put the changes through an adversarial pass with independent harnesses. That found one blocker and a cluster of false-positive gaps. Everything below was reproduced against the PR head (d7dcf20); numbers are from those runs.

## Blocker: the filename mask at `detector/patterns.js:771` has cubic backtracking, reachable through `analyzeText`

`\b[a-z0-9_.-]+-[a-z0-9_.-]+\.[a-z0-9]{1,16}\b` shares `-` and `.` between two adjacent unbounded quantifiers, then requires a `\.` that never arrives on a long kebab run. Measured growth is almost exactly cubic (3.37x / 2.38x / 2.27x per step against cubic predictions of 3.375 / 2.37 / 2.28):

| kebab blob | total doc | PR branch | main |
|---|---|---|---|
| 6.0 KB | 8.1 KB | 5,592 ms | 1.6 ms |
| 9.0 KB | 11.1 KB | 18,854 ms | 1.8 ms |
| 12.0 KB | 14.1 KB | 44,917 ms | 2.3 ms |
| 15.8 KB | 17.9 KB | 101,815 ms | 2.9 ms |

The word-count guards don't help: a pasted slug list or generated identifier is one "word" (`countWords` splits on `\S+`), and there is no character-length guard. The path mask at line 770 is independently quadratic (~4x per doubling), so fixing 771 alone leaves the function superlinear. Bounding the quantifiers on both lines (suggestion on the diff) measured linear, took the worst case from 5,470 ms to 52 ms through `analyzeText`, and produced zero behavior changes across the masking probes — `my-file.md`, `foo-bar.test.js`, `some_file-v2.tar.gz` all still masked. Atomic groups and possessive quantifiers aren't supported in JS regex, so bounding is the right shape.

One ask beyond the fix itself: a regression test with a timing assertion on the worst-case input. A correctness-only fixture passes both before and after, so it guards nothing here.

## False positives — each fires today with a rewrite that is wrong

1. **Proper nouns and identifiers.** The compound patterns are `/gi` and the suggestion is a plain hyphen-drop, so capitalization survives into rewrites that are never correct English: `Code-Base Enterprise` → `CodeBase`, `Road-Map Analytics` → `RoadMap`, a `CODE-BASE` heading → `CODEBASE`. Kebab identifiers in prose (npm package names, CSS classes, config keys) fire too. Your own spec item 1 in #107 lists identifiers and version strings among the protected spans; that masking never got implemented, and it is the root of this class. For the proper-noun remainder, not firing on capitalized compounds is probably the cheapest correct rule.
2. **Flags with punctuation prefixes.** `(?<!\S)` masks bare `--code-base` but `(--code-base)`, `,--code-base`, `[--code-base]`, and `=--code-base` all fire. Backticked flags are fine, which limits the blast radius in well-formatted docs, but changelogs and plain-prose docs hit this.
3. **Single-component paths.** Multi-component paths and dotpaths are correctly masked (`./code-base`, `src/code-base`, `C:\fixtures\code-base` all quiet), but `C:\code-base`, `C:/code-base`, `~/code-base`, `/code-base`, and trailing-slash `code-base/` all fire. The must-not-fire fixture uses a two-component path — exactly the neighbor that passes; every single-component neighbor fails.
4. **Metadata is not prose.** YAML frontmatter (titles, tag lists), Markdown table cells, unfenced YAML blocks, and unquoted HTML attributes all fire. `check-style.js` already masks frontmatter, so there is repo precedent to borrow.

Please add must-fire and must-not-fire fixtures for the failing neighbors above as part of the fixes — the current fixtures cover the passing neighbors only.

## Classification drift in short documents

Pure copyedits should not move the trinary. At roughly 100 words or fewer, five curated compounds move the score 0 → 10 and the label from "Clean" to "Minimal AI signals" (human probability 1.0 → 0.891); eight distinct forms in a 48-word document flip `HUMAN_ONLY → MIXED` (human 1.0 → 0.459) with zero non-hyphenation issues, through the `score >= 15` path. The inline comment's "never act as a trinary classifier corroborator" holds for the corroborator lists, but this path routes around it. We solved this exact problem for `tier1-clarity` with weighting — same treatment here, please. Longer documents are fine (score 4 at ~1,200 words); it's the short-doc case that needs the guard.

## On the Qodo review above

Finding 1 is wrong — keep your CLAUDE.md and CHANGELOG counts exactly as you have them. The CLAUDE.md copy is *required*, not forbidden: `scripts/check-pattern-count.sh` hard-asserts it against SKILL.md (I ran the control pair — reverting your bump to 61 fails CI with `pattern-count drift`), `.ssot.yaml` registers CLAUDE.md as a tracked copy, and dated CHANGELOG entries have stated counts in every release this year. Finding 2 is real (item 2 above). Finding 3 names the wrong gap — dotpaths work; single-component paths are the hole (item 3 above).

## Before this lands

The bounded regexes on 770-771 plus the timing-asserted regression test, identifier/version-string masking per your #107 spec, the flag and path mask fixes with fixtures for the failing neighbors, frontmatter/table masking, and the short-doc weighting guard. It's a real list, but nothing structural — the curated-subclass approach and the fixture discipline are exactly right, which is why I'm confident the fixes will hold. CI hasn't run yet (first-contribution gate); I'll approve the workflow run when you push.

conorbronsdon:
Making a small addition but otherwise looks good thanks @mbhall88

detector/patterns.js
```diff
@@ -709,23 +719,195 @@ const AIDetector = (() => {
...
+    maskMatches(/^[ \t]*>[^\n]*$/gm);
+    maskMatches(/\b(?:https?:\/\/|www\.)[^\s<>]+/gi);
+    maskMatches(/<[!?/]?[a-z][^>\n]*>/gi);
+    maskMatches(/(?<![a-z0-9_-])--?[a-z0-9][a-z0-9-]{0,127}/gi);
+
+    // Paths and filenames use bounded components. Besides preventing
+    // superlinear backtracking on long kebab blobs, the explicit prefix and
+    // trailing-slash forms cover single-component paths such as C:\\code-base,
+    // ~/code-base, /code-base, and code-base/.
+    maskMatches(/(?:[a-z]:[\\/]|\.{1,2}[\\/]|~[\\/]|[\\/])[a-z0-9_.-]{1,64}/gi);
```
conorbronsdon: The current 64-character bound masks only the prefix of a valid longer component, leaving `code-base` visible after character 64.

```suggestion
    maskMatches(/(?:[a-z]:[\\/]|\.{1,2}[\\/]|~[\\/]|[\\/])[a-z0-9_.-]{1,255}/gi);
```

detector/patterns.test.js
```diff
@@ -355,6 +355,245 @@ test('"load-bearing" (metaphor) flags tier1; construction nouns exempt', () => {
...
+    'Pass ,--code-base when starting the local compatibility service for older clients.',
+    'Use [--code-base] when documenting the local compatibility service for older clients.',
+    'Set mode=--code-base when starting the local compatibility service for older clients.',
+    'The Windows file lives at C:\\code-base for users of the legacy client.',
+    'The portable file lives at C:/code-base for users of the legacy client.',
+    'The home-directory file lives at ~/code-base for users of the legacy client.',
+    'The root-level file lives at /code-base for users of the legacy client.',
+    'The generated output is copied into code-base/ during every local release build.',
+    'The relative file lives at ./code-base for users of the legacy client.',
+    'The parent-relative file lives at ../code-base for users of the legacy client.',
```
conorbronsdon, line 502: These fixtures pin the Unix and Windows forms that currently leak from a 65+ character component.

```suggestion
    'The parent-relative file lives at ../code-base for users of the legacy client.',
    `The generated file lives at /${'a'.repeat(65)}-code-base for users of the legacy client.`,
    `The generated Windows file lives at C:\\${'a'.repeat(65)}-code-base for legacy users.`,
```

conorbronsdon (commented):
One small follow-up from final validation: a single path component longer than 64 characters can still leak a curated compound from the unmasked suffix. I reproduced it with both `/` and `C:\\` prefixes plus a 65-character component ending in `-code-base`; each returns an incorrect `unnecessary-hyphenation` finding.

The two suggestions below keep the regex bounded at the common 255-character filesystem component limit and add Unix/Windows regression coverage. I applied the same patch locally and confirmed `npm test`, `npm run self-scan:check`, `scripts/check-pattern-count.sh`, and the explicit long-path probes all pass.

conorbronsdon (approved):
Thanks for your work on this!

PR #127, opened 2026-08-13
fix: require blank closing code fences

### LINKED ISSUE #77: fenceRanges() accepts an info string on a closing fence, ending code blocks early
## Summary

`fenceRanges()` in `detector/patterns.js` (added in 3b5bbc11, the #62 fix)
treats a fence-like line *with an info string* as a valid closing fence.
CommonMark permits an info string on an opening fence but not on a closing
one, so an open code block can be closed early, and content still inside the
block gets scanned.

## Detail

The scanner is:

```js
const re = /^[ \t]{0,3}(`{3,}|~{3,})[^\n]*$/gm;
...
} else if (marker[0] === open.char && marker.length >= open.len) {
  // treated as the closer
```

The `[^\n]*` accepts arbitrary trailing text, and the closing branch checks
only the delimiter character and length. It never requires the remainder of a
*closing* line to be blank.

So in this document the second line is taken as the closer:

````markdown
```
```js
## Benefits And Strategic Considerations
```
````

The heading is still inside the code block, but because the fence range ended
early it is scanned and emitted as a `title-case-header` issue.

## Suggested fix

When deciding whether a matched line closes an open fence, require the text
after the delimiter to be whitespace-only. Capturing the remainder and testing
it is enough, e.g.:

```js
const re = /^[ \t]{0,3}(`{3,}|~{3,})([^\n]*)$/gm;
...
} else if (marker[0] === open.char && marker.length >= open.len
           && m[2].trim() === '') {
```

Openers keep accepting an info string; only closers become strict.

## Context

Found while refreshing a downstream vendored snapshot to v3.22.1. We keep
`detector/patterns.js` byte-identical to a pinned commit, so we are reporting
rather than patching locally.

mjq2020:
## Summary

- capture the text after a fenced-code delimiter
- only allow a matching delimiter with whitespace-only trailing text to close an open fence
- cover an info-string fence nested inside an outer fenced block

## Root cause

`fenceRanges()` accepted every fence-like line as a possible closer, including lines such as `````js``. CommonMark allows an info string on an opening fence, but a closing fence may only be followed by spaces or tabs. The premature close exposed headings inside the outer code block to prose detectors.

## Tests

- `npm test`
- `npm run self-scan:check`
- `bash scripts/check-pattern-count.sh`

Fixes #77

### DIFF
--- a/detector/patterns.js
+++ b/detector/patterns.js
@@ -693,7 +693,7 @@ const AIDetector = (() => {
       } else if (
         marker[0] === open.char &&
         marker.length >= open.len &&
-        m[2].trim() === ''
+        /^[ \t]*\r?$/.test(m[2])
       ) {
         ranges.push([open.start, m.index + m[0].length]);
         open = null;
--- a/detector/patterns.js
+++ b/detector/patterns.js
@@ -682,7 +682,7 @@ const AIDetector = (() => {
    * sliced the whole document for every candidate, which is quadratic on a
    * heading-dense file. */
   function fenceRanges(text) {
-    const re = /^[ \t]{0,3}(`{3,}|~{3,})[^\n]*$/gm;
+    const re = /^[ \t]{0,3}(`{3,}|~{3,})([^\n]*)$/gm;
     const ranges = [];
     let open = null;
     let m;
@@ -690,7 +690,11 @@ const AIDetector = (() => {
       const marker = m[1];
       if (!open) {
         open = { char: marker[0], len: marker.length, start: m.index };
-      } else if (marker[0] === open.char && marker.length >= open.len) {
+      } else if (
+        marker[0] === open.char &&
+        marker.length >= open.len &&
+        m[2].trim() === ''
+      ) {
         ranges.push([open.start, m.index + m[0].length]);
         open = null;
       }
--- a/detector/patterns.test.js
+++ b/detector/patterns.test.js
@@ -1282,6 +1282,18 @@ test('#62: fences that a parity count gets wrong', () => {
   );
 });
 
+test('#77: a fence with an info string cannot close an outer fence', () => {
+  const f3 = '```';
+  const intro = 'Documentation about writing Markdown, long enough to clear the word gate.';
+  const title = '## Benefits And Strategic Considerations';
+
+  assert.equal(
+    titleCaseHits([intro, f3, f3 + 'js', title, f3, f3].join('\n') + HEADING_BODY).length,
+    0,
+    'an info string belongs to an opening fence, so the heading remains code',
+  );
+});
+
 test('#62: MD_HEADING_PREFIX accepts what the pattern accepts', () => {
   // The two must stay coupled: TITLE_CASE_HEADER matches `#{1,6}[ \t]+`, so if
   // the prefix strip stops accepting a tab, `##` survives into the token list
--- a/detector/patterns.test.js
+++ b/detector/patterns.test.js
@@ -1282,16 +1282,32 @@ test('#62: fences that a parity count gets wrong', () => {
   );
 });
 
-test('#77: a fence with an info string cannot close an outer fence', () => {
-  const f3 = '```';
+test('#77: only spaces and tabs may follow a closing fence', () => {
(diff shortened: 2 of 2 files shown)


conorbronsdon:
Thanks for your submission! Will take a look

detector/patterns.js
```diff
@@ -682,15 +682,19 @@ const AIDetector = (() => {
...
     let m;
     while ((m = re.exec(text)) !== null) {
       const marker = m[1];
       if (!open) {
         open = { char: marker[0], len: marker.length, start: m.index };
-      } else if (marker[0] === open.char && marker.length >= open.len) {
+      } else if (
+        marker[0] === open.char &&
+        marker.length >= open.len &&
+        m[2].trim() === ''
```
conorbronsdon: `trim()` also removes characters such as a non-breaking space, which CommonMark does not allow after a closing fence. The optional `\r` preserves CRLF input because the capture includes the carriage return.

```suggestion
        /^[ \t]*\r?$/.test(m[2])
```

detector/patterns.test.js
```diff
@@ -1282,6 +1282,18 @@ test('#62: fences that a parity count gets wrong', () => {
...
+  const f3 = '```';
+  const intro = 'Documentation about writing Markdown, long enough to clear the word gate.';
+  const title = '## Benefits And Strategic Considerations';
+
+  assert.equal(
+    titleCaseHits([intro, f3, f3 + 'js', title, f3, f3].join('\n') + HEADING_BODY).length,
+    0,
+    'an info string belongs to an opening fence, so the heading remains code',
+  );
+});
```
conorbronsdon: Please cover both fence characters and the exact suffix boundary: spaces/tabs and CRLF may close; NBSP and other trailing text may not. I verified that the NBSP assertion fails with the current `trim()` predicate and passes with the stricter check.

```suggestion
test('#77: only spaces and tabs may follow a closing fence', () => {
  const intro = 'Documentation about writing Markdown, long enough to clear the word gate.';
  const title = '## Benefits And Strategic Considerations';

  for (const fence of ['```', '~~~']) {
    assert.equal(
      titleCaseHits([intro, fence, fence + 'js', title, fence].join('\n') + HEADING_BODY).length,
      0,
      'trailing text cannot close the outer fence',
    );
    assert.equal(
      titleCaseHits([intro, fence, 'code', fence + ' \t', title].join('\n') + HEADING_BODY).length,
      1,
      'spaces and tabs may follow a closing fence',
    );
    assert.equal(
      titleCaseHits([intro, fence, 'code', fence + '\u00a0', title].join('\n') + HEADING_BODY).length,
      0,
      'non-breaking space is fence content, not an allowed closing-fence suffix',
    );
    assert.equal(
      titleCaseHits([intro, fence, 'code', fence, title].join('\r\n') + HEADING_BODY).length,
      1,
      'CRLF line endings still allow a closing fence',
    );
  }
});
```

conorbronsdon (changes requested):
Thanks for the focused fix — the reported case is handled and the approach is right. One narrow CommonMark boundary remains: `String#trim()` accepts Unicode whitespace beyond the spaces and tabs allowed after a closing fence. Please tighten that predicate and add the boundary fixtures below. I’ll handle the changelog/versioning in the next batched release PR, so no release-metadata change is requested here.

conorbronsdon:
@mjq2020 let me know if you have any questions, otherwise think we'll be good after edits above are applied

detector/patterns.js
mjq2020: Addressed in 23d46b1: the closer suffix now uses `/^[ \\t]*\\r?$/`, so only spaces/tabs (plus the CR captured from CRLF input) are accepted.

detector/patterns.test.js
mjq2020: Addressed in 23d46b1. The regression test now covers both backtick and tilde fences, spaces/tabs, NBSP, trailing info text, and CRLF.

conorbronsdon (approved):

conorbronsdon:
Thanks; appreciate the contribution and engagement!

Issue #108, opened 2026-08-04
six patterns my merged system flags that aren't in v3.23


odinfree:
took you up on the offer. i ran my merged catalog against SKILL.md v3.23.0 and pulled out everything mine flags that yours doesn't. six patterns, written close to your entry format so you can lift what's additive. examples are the shapes i keep deleting from drafts in my genre (crypto/ZK gtm, daily posting).

**1. false agency**

active-voice sentences that hand human judgment, intent, or responsibility to an inanimate subject.

- "the data tells us adoption is early"
- "the market rewards shipping"
- "a bet lives or dies on distribution"
- "the decision emerged after the offsite"
- "the culture shifted toward accountability"

why the current rules miss it. agentless passives catches the hidden subject ("support was added"). moral-adjective category errors catches the adjective on a non-agent ("an honest shape"). this is the third seat, subject present, voice active, but the actor is a thing that can't decide. each sentence reads fine until you ask who did it, and then the responsibility has no owner.

carve-out. ordinary literal verbs stay. a report can show a result, a system can return an error, a paper can argue a position. flag only when the verb carries judgment, intent, choice, or responsibility that some person exercised.

fix. name the human. "the decision emerged after the offsite" becomes "we decided after the offsite." "the data tells us adoption is early" becomes "the funnel numbers read early to us."

**2. the fake-casual kit**

the register models emit when asked for a lowercase casual social voice. you already catch "Plot twist:" (infomercial hooks) and "Real talk:" / "Honestly?" (the fake-candid add-on). the rest of the kit is missing, and it's the part that survives cleanup because it sits closest to an actual casual voice:

- one-word verdict closers as the whole closing line. "wild." "insane." "unhinged."
- stage directions. "\*checks notes\*" "\*chef's kiss\*" "\*mic drop\*"
- wink asides. "(yes, really)" "(no, seriously)"
- label-prefix openers beyond plot twist. "hot take" "fun fact" "pro tip" "PSA" "unpopular opinion", with or without the colon
- "because of course it does"
- the self-QA volley. "Is it fast? Yes. Is it cheap? Also yes."

the tell across all six props is that the drama is outsourced to the prop instead of carried by the content. a post can clear every vocabulary tier and still be wearing this costume, which is exactly why it slips through.

fix. delete the label and say the thing. replace the verdict word with the specific surprise. cut the wink and the stage business. a strong observation needs no costume.

**3. launch-copy dramatic introductions**

- "Enter X."
[...]

PR #128, opened 2026-08-16
Add detector trio from #108: launch intros, crowd contrast, fake-casual props

odinfree:
## Summary

the detector trio from #108, first half of the split you asked for. three new `type`s, each with must-fire and must-not-fire fixtures:

- `launch-intro`: "Enter X.", "Meet X, your new favorite...", "Think X meets Y", gated to sentence-initial imperatives with a capitalized product name. the Meet surface additionally requires a launch-copy head ("your new favorite", "the new standard") so human introductions ("Meet Sarah, your new account manager", shelter posts, baby announcements) stay clean. "Say hello to X" stays judgment-only. known accepted FPs, disclosed in the entry: dramatic scripts ("Enter Hamlet.") and column narrative ("Enter Rashford.").
- `crowd-contrast`: the trailing "while everyone else..." clause, gated on a dismissive verb plus the "was still" dramatization marker. wire copy, memoir, and fiction using those verbs literally stay clean ("others debated the amendment" does not fire). verb stems carry explicit inflection tails so "market speculators" and "deliberately ignored" never match, and the accepted residue (a literal "was still debating") is disclosed. recall is deliberately sacrificed: "was busy debating" without "still" stays a miss.
- `fake-casual-prop`: stage directions (including the curly-apostrophe *chef's kiss* form LLMs actually emit), wink asides, and "because of course it does" gated to present tense plus "did", so past-tense human grumbles ("because of course it was raining") stay clean. verdict closers and label-prefix openers stay skill-only.

beyond the corpus gate: 18 constructed probes across news wire, hr intros, install docs, memoir, and sports narrative, all correct, with the must-not-fire set pinned as fixtures. fp-measure against corpus/human/: zero hits from all three types on 875 human paragraphs. catalog 62 to 65, engine 48 to 51, word table unchanged at 112. the catalog entries follow in a second pr stacked on this branch.

## Checklist

- [x] If I added a detector `type`: it's documented in `detector/CATEGORIES.md` and has a fixture in `detector/patterns.test.js` (a true positive **and** a must-not-fire case)
- [ ] If I added a judgment-only rule: n/a in this pr (those are the stacked second pr)
- [x] Any factual claim about how AI or humans write cites a source
- [x] The prose I added passes the skill's own audit (no AI-writing tells, terse bullets, no hollow intensifiers)
- [x] `CHANGELOG.md` entry added under a dated `## [X.Y.Z]` heading, and `SKILL.md` `version:` bumped


conorbronsdon (changes requested):
The split and the precision-over-recall approach are right, but three detector surfaces need tightening before merge:

1. Move bare `Enter X.` to judgment-only. The current branch matches ordinary UI/docs copy including `Enter Password.`, `Enter Amount.`, and `Enter Username — your email.`. Removing dash terminators or adding a field-name denylist will not solve the period-terminated class. Please keep the safer `Meet X, your new favorite…` and `Think X meets Y` surfaces and pin the UI cases as must-not-fire fixtures.

2. Make the first crowd-contrast branch actually progressive-only. After `was/were/is/are still`, the regex currently accepts all inflections, so it matches `was still deliberate about…`, `was still debated by pundits`, and even `was still debates timelines`. Restrict that branch to `-ing` forms and add adjective/passive negative fixtures.

3. Reconsider deterministic detection of `because of course …`. Including `did`, `is`, and `are` flags ordinary human grumbles such as “The build failed because of course it did.” The present-vs-past rationale in the comment does not hold when `did` is included. Under this repo’s precision policy, I would move this surface to judgment-only; a stronger contextual/co-occurrence gate with adversarial negatives would also be acceptable.

Please also remove the catalog total from the changelog and make the CLAUDE.md single-source instruction count-agnostic—it currently says the README is the only source while repeating the number itself.

I am intentionally **not** requesting Qodo’s “Say hello” detector or exhaustive detection of verdict closers, label prefixes, and self-QA. Those require voice/context judgment and deterministic coverage would create more false positives. After these changes, please run the corpus and full test suite once the fork workflows are approved; the current Actions runs show `action_required` and created no jobs.

conorbronsdon:
Correction to one subpoint in my review after checking `scripts/check-pattern-count.sh` and `.ssot.yaml`: please **keep updating the numeric count in CLAUDE.md**. It is an intentional, CI-enforced tracked copy of the README value. I withdraw the request to make that sentence count-agnostic.

The changelog point is narrower: the current written guidance says not to restate the current catalog total elsewhere, although historical changelog entries already preserve release snapshots. Either remove the new changelog total (smallest change) or clarify the guidance that version-scoped changelog snapshots are allowed. This is documentation-policy cleanup, not a detector correctness blocker.

conorbronsdon:
One additional documentation mismatch from the adversarial pass: the PR body, SKILL entry, and CATEGORIES mapping say the crowd-contrast detector requires a dismissive verb **plus `was still`**, but only the debate/speculation branch has that gate. The `writing/wrote thinkpieces` and `play…catch-up` branches intentionally match without it. Please scope the prose claim to the progressive debate/speculation branch and document the two directly detected stereotyped variants, rather than implying one universal gate.

conorbronsdon:
Cleaning this up for merge

conorbronsdon:
Thanks for this, @odinfree. Rather than round-trip again I pushed four maintainer commits onto your branch (your commit is untouched; the squash keeps you as author):

- Merged current `main` (the #138 categories landed in the same regions), so this is now 3.29.0 with 72 pattern categories and 54 engine types, plugin and Cursor copies regenerated.
- The review items: bare `Enter X.` and `because of course …` are judgment-only now, the first crowd-contrast branch is progressive-only with the adjective/passive negatives pinned, the "dismissive verb + was still" claim is scoped to the branch that has it, and the changelog entry carries no totals.
- Two more precision fixes from a second review pass: `the new home` / `the new way` no longer swallow compound nouns ("Meet Rosa, the new home secretary" is clean; "the new home of" still fires), and `*chefs kiss*` without an apostrophe no longer matches.
- Each entry now states exactly what its regex does, including what it deliberately misses (two-token names, crowds outside the closed list) and the person-name residue that still fires.

Full suite, self-scan, pattern-count gate, and both sync scripts pass locally; eight mutation controls confirm the new fixtures can fail. Merging once CI agrees.

conorbronsdon (approved):

PR #130, opened 2026-08-16
Scope repeated reversal punchlines to hollow repetition

cland4449:
## Summary

Complete the contributor's reversal-pattern proposal as a P2, judgment-only subtype of “Manufactured punchlines and staccato drama.” Flag repeated reversals only when they replace concrete claims. Supported contrasts, isolated intentional lines, comedy, fiction, speeches, and quotations pass. Rewrites preserve source facts and ask for missing explanations.

Integrate current main while retaining the original contributor commit. Synchronize version 3.30.0 and all generated bundles; preserve the record of the published 3.29.0 plugin. The detector and catalog count remain unchanged.

## Validation

- `npm test`, count/sync checks, plugin and router validators, preservation smoke, and deterministic packaging pass. Windows skipped the symlink-negative test because the account lacks symlink privileges.
- Reviewed commit `081b495614e617b273e6eb7a0217d60f13622a56` with Claude Sonnet 5, paid Go Grok 4.6, and free Muse Spark 1.3 and Nemotron 3.5 Lightning. No verified blockers remain. Claude, Grok, and Muse matched all eight semantic cases; Nemotron's one context-related false positive was checked and rejected against the explicit passage-level carve-out.

## Checklist

- [x] Tests and category contract pass.
- [x] Judgment-only subtype documented in `detector/CATEGORIES.md`; no detector type added.
- [x] Flag/pass examples and legitimate-writing carve-outs included.
- [x] No unsupported AI-prevalence claim or invented rewrite details.
- [x] Dated changelog entry, version metadata, and generated copies synchronized.


conorbronsdon (changes requested):
Thanks for this contribution. I think your underlying observation is valid, but the rule needs a narrower operative definition.

There is meaningful evidence that LLMs overuse the broader family of contrastive reframing and manufactured reversals. The Washington Post found recurring “not just X, but Y” constructions in an analysis of 328,744 ChatGPT messages, while Pangram reports that this form appears roughly three times as often in AI text as in human text. Writers in OpenAI’s community independently describe the same cluster: opposing sentence pairs, clipped reversals, staccato paragraphs, and closers that land as generic one-liners.

Sources:

- https://www.washingtonpost.com/technology/interactive/2025/how-detect-chatgpt-em-dash/
- https://www.pangram.com/supporting-evidence
- https://community.openai.com/t/most-annoying-habit-can-i-make-it-stop/1265708

I could not find comparable evidence identifying paraprosdokians as a whole as an LLM signature. It is a broad, longstanding human rhetorical device, particularly in comedy, speeches, fiction, and advertising. The research supports repetition and formulaic deployment as the signal, not the mere presence of a reversal. I do however, anecdotally agree that paraprosdokians are currently an indicator of LLM writing, even if there isn't clear research (please correct me if I missed something).

I’d like to keep the insight while narrowing the rule:

1. Rename it to something like “Repeated setup→reversal punchlines.” “Paraprosdokian” can remain as an explanatory term, but the current heading makes the entire rhetorical device sound suspect, which I think is over the top for the moment 

2. Make repetition or hollowness part of the operative test. Flag when:
   
   - two or more reversals appear in one piece;
   - they cluster in hooks, closers, or final list items;
   - the reversal substitutes for a concrete claim; or
   - they co-occur with negative parallelism, staccato fragments, or other manufactured-wit patterns.

3. Do not classify one supported, voice-appropriate reversal as P1. Either make the category P2, or require repetition/stacking before the P1 trigger fires. An isolated instance should normally pass.

4. Replace the categorical claim that “LLMs deploy it as a default closer” with something supportable, such as:
   
   «Readers increasingly perceive repeated contrast-and-reversal structures as AI-like. One occurrence is ordinary human rhetoric; repetition or using the reversal in place of specific information creates the signal.»

5. Add explicit pass examples so reviewers can distinguish the device from its overuse. A reversal that communicates a concrete, supported distinction should pass, along with intentional comedy, quotations, speeches, fiction, and perhaps a single voice-appropriate line.

6. Please revise the proposed fix that introduces disk, network, and clock-skew details. Those facts are not present in the source sentence, so the rewrite risks inventing information. If the missed failure mode is unknown, the safe fix is to state that plainly or ask the writer to supply it.

I’m open to this remaining a standalone category if its boundary with “Manufactured punchlines and staccato drama” is made explicit. Folding it into that existing category as a documented subtype would also be reasonable.

The remaining mechanical items should be handled after rebasing onto current "main":

- add the judgment-only category to "detector/CATEGORIES.md";
- synchronize all version metadata, including "package.json";
- update the stale PR title/body, version, and catalog counts;
- remove or reconcile the extra changelog count statement under the repository’s current count policy.

So this is a request for changes, but not a rejection of the observation. I think you have identified a real reader-perception problem; the revision should target repeated, hollow reversal cadence rather than treating the underlying rhetorical device as inherently AI-like imo

conorbronsdon:
Thanks @cland4449 for the contribution!

PR #131, opened 2026-08-21
docs: document npx skills as a universal install method

mbhall88:
## Summary

- Adds a "Quick install — any agent" section at the top of Installation & Usage documenting [`npx skills add`](https://github.com/vercel-labs/skills) (Vercel Labs' community CLI for the open agent-skills ecosystem) as the fastest way to install this skill, with `npx skills update` covering future updates.
- Pins the documented commands to `skills@1.5.23` (not `@latest`) so the installer version that runs is fixed and visible, while clarifying that the skill payload still follows this repository's current default branch. Also notes the CLI's Node `>=22.20.0` requirement.
- Accurately describes the CLI's behavior against this public repository: the normal GitHub blob fast path for this root-level skill installs only `SKILL.md` ([source](https://github.com/vercel-labs/skills/blob/v1.5.23/src/blob.ts#L619-L641)), but an unavailable blob path can trigger a clone fallback that installs the full root skill directory ([regression test](https://github.com/vercel-labs/skills/blob/v1.5.23/tests/root-level-disk-install.test.ts)). `git clone` remains the documented route when the detector, tests, and other repo tooling are explicitly wanted. `skills update` is described as refreshing the selected scope (it prompts project/global; `-g`/`-p` for non-interactive), not every install at once.
- Adds `metadata.repository` to `SKILL.md`'s frontmatter, so a copy installed via the normal SKILL.md-only fast path still carries a link back to the project and its contributor community, alongside the existing `author` field. Synced to the plugin copy via the existing `sync-plugin-skill.sh`.
- Existing manual per-platform steps (Claude Code, Cowork, OpenClaw, Cursor, Hermes, Codex, etc.) are unchanged and kept as a no-Node fallback.

Ships as `3.25.2` with a CHANGELOG entry — this repo's conventions expect a dated entry and version bump for a notable change even without a rule/detector change, and I originally opened this as `3.25.1` before upstream `main` picked up that version number from #133 in the meantime; this PR now rebases past that and re-bumps to `3.25.2` to avoid the collision.

### Why `skills` instead of a bespoke installer

I initially built a standalone `avoid-ai-writing` npm package/CLI for this, but `npx skills add` already solves it end-to-end, and is a better fit than anything this repo would realistically maintain across 75+ agent harnesses.

## Test plan

- [x] `npm test` — passes
- [x] `npm run self-scan:check` — README stays within its AI-writing-pattern budget after the new prose
- [x] `bash scripts/check-pattern-count.sh` — counts unaffected
- [x] `bash scripts/sync-plugin-skill.sh && bash scripts/sync-cursor-rules.sh` — no drift (regenerated copies match, including the new `repository` frontmatter field)
- [x] Verified `npx skills@1.5.23 add conorbronsdon/avoid-ai-writing -a claude-code -y` against the real upstream repo: the normal public-GitHub blob path installed only `SKILL.md`; this verifies that normal path, not the clone fallback.


README.md
```diff
@@ -42,10 +42,38 @@ A one-shot "make this sound human" prompt catches the obvious stuff. This skill
...
-- **Works across platforms** — one `SKILL.md` runs in Claude Code, Cowork (as a plugin), OpenClaw, and Cursor (as a ported rule). See the install paths below.
+- **Works across platforms** — one `SKILL.md` runs in Claude Code, Cowork (as a plugin), OpenClaw, Cursor (as a ported rule), and more via `npx skills add`. See the install paths below.
 
 ## Installation & Usage
 
+### Quick install — any agent
+
+The fastest way to install this skill, and later keep it updated, is the community [`skills`](https://github.com/vercel-labs/skills) CLI. It auto-detects which agents you have installed and supports 75+ of them, including every one listed below. `npx` downloads and runs that package from the npm registry — it's third-party code, not something this repo publishes or maintains, so review it if you'd rather not run an unpinned package: the `@latest` below is explicit so the version in play is visible, and you can swap in a specific version (e.g. `skills@1.5.23`) for a reproducible install.
```
conorbronsdon: `@latest` is still a moving dist-tag, so it does not make the resolved version visible or the install reproducible. Please describe that accurately, retain `skills@1.5.23` as the pinned example, and note the CLI’s current Node requirement (`>=22.20.0`). The third-party-code warning itself is useful.

README.md
```diff
@@ -42,10 +42,38 @@ A one-shot "make this sound human" prompt catches the obvious stuff. This skill
...
+
+```bash
+npx skills@latest add conorbronsdon/avoid-ai-writing
+```
+
+That pulls the whole repo into the skill directory (matching the git-clone method below). To install just `SKILL.md` instead:
```
conorbronsdon: This whole-repo claim is not reliable with the current CLI. Its public-GitHub blob fast path deliberately filters a root-level skill to `SKILL.md` only ([upstream implementation](https://github.com/vercel-labs/skills/blob/main/src/blob.ts#L633-L641)); only a fallback path may copy the full repo. Please say this installs the root skill and keep `git clone` as the documented route for the detector, configs, and other repository tooling.

README.md
```diff
@@ -42,10 +42,38 @@ A one-shot "make this sound human" prompt catches the obvious stuff. This skill
...
+
+```bash
+# Target specific agents instead of everything detected
+npx skills@latest add conorbronsdon/avoid-ai-writing -a claude-code -a codex
+
+# Install globally (~/.<agent>/skills/) instead of the current project
+npx skills@latest add conorbronsdon/avoid-ai-writing -g
+```
+
+Later, `npx skills@latest update` refreshes every skill installed this way — see the [`skills update` docs](https://github.com/vercel-labs/skills#skills-update) for the full command reference.
```
conorbronsdon: `skills update` prompts for project/global scope, so it does not necessarily refresh every installed skill across both scopes. Please say it refreshes installed skills in the selected scope (with `-g` / `-p` available for an explicit scope).

conorbronsdon (changes requested):
This is the right direction: a broadly compatible installer lowers adoption friction while preserving attribution through the `conorbronsdon/avoid-ai-writing` source identity. Please make the focused accuracy fixes in the inline comments, then update the PR description/test plan to match the current 3.25.1 version + changelog diff. I’m intentionally keeping the “any agent” positioning.

conorbronsdon:
One attribution/provenance follow-up while making these changes: the `skills` fast path can leave users with only `SKILL.md`. That file names `author: Conor Bronsdon` but contains no source URL. Please add a portable repository/homepage field in its metadata (for example `repository: https://github.com/conorbronsdon/avoid-ai-writing`) so copies retain a direct path back to the project and its contributor community. The generated plugin copy should stay in sync through the existing script.

README.md
mbhall88: Fixed: reverted to a pinned `skills@1.5.23` (dropped `@latest`) and added the Node `>=22.20.0` requirement note alongside the third-party-code warning.

README.md
mbhall88: Fixed: rewrote this to say a root-level skill installs `SKILL.md` only via the CLI's blob fast path (linked src/blob.ts as the source), not the whole repo, and pointed to `git clone` for the detector/tests/CI config. Also dropped the now-redundant raw-`SKILL.md`-URL example since the shorthand does the same thing.

README.md
mbhall88: Fixed: now says it refreshes installed skills in whichever scope you select (prompts project/global, or pass -g/-p non-interactively) rather than "every skill installed this way."

mbhall88:
Added `metadata.repository: https://github.com/conorbronsdon/avoid-ai-writing` to SKILL.md's frontmatter, alongside the existing `author` field, so an SKILL.md-only install still links back to the project. The plugin copy stays in sync through `sync-plugin-skill.sh` as before. All three inline comments are also fixed — replied on each. Pushed as a new commit; `docs/npx-skills-install` is rebased past main (which picked up 3.25.1 via #133 while this was open) and now ships as 3.25.2.

conorbronsdon (approved):
Validated at 211689ee329545db4ed65a6ed524809038f69e40. The installer pin and root-skill fallback wording now match the documented v1.5.23 behavior. Local validation and both required CI checks pass; ready to merge.

conorbronsdon:
Thanks for the contribution @mbhall88! I appreciate it

PR #133, opened 2026-08-22
Bind voice-profile targets to Never-inject guardrails (fixes #100)

### LINKED ISSUE #100: Voice profile targets can mandate content the source lacks, colliding with the never-inject guardrails
Surfaced by CodeRabbit on the wshobson/agents port (wshobson/agents#645): three voice-profile targets mandate content the source may not contain, which collides with the "Never inject these" guardrails when the source lacks the target.

- `casual`: "At least one first-person or concrete-anecdote touch" vs "Fake first person: ... If the source has no `I`, the rewrite has no `I`."
- `professional`: "One concrete claim per paragraph (a number, a name, a date)" and "Make the ask explicit" vs "Invented specifics: ... If the concrete detail is missing, flag the gap and leave it. Never fill it."
- `warm`: "acknowledge them at least once" can require inventing an acknowledgment.

The Never-inject section wins by its own terms ("None of the following may be added"), but the profiles never say so, and a model following the targets checklist-style will fabricate the missing touch. The port resolved it by bounding each target on source evidence, e.g. casual became "Keep first-person and concrete touches where the source has them; never add one it lacks", professional became "Prefer a concrete claim ... when the source provides one. Keep the ask explicit where the source makes one; never invent facts or an ask", warm became "address the reader directly where the source already speaks to them, and keep its acknowledgment rather than adding one", plus one intro line stating the rewrite guardrails bind voice targets.

Fixed in the port in wshobson/agents#645 commit f78a92e; the SSOT §Voice profiles still has the unconditional wording.

mahinNadir:
Fixes #100. casual, professional, and warm each had a target that could only be satisfied by adding content the source lacked, colliding with the Never-inject guardrails. Reworded each target to apply only where the source already has the relevant material, and added one line stating the guardrails bind voice targets. Same resolution already validated downstream in wshobson/agents#645 (commit f78a92e), ported back to the SSOT here. Tests: npm test passes (39/39 in check-style, full suite green).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project are documented here.
 
 ---
 
+## [3.25.1] — 2026-08-21
+
+### Fixed
+
+- **Voice-profile targets are bound to the Never-inject guardrails (#100).** `casual`, `professional`, and `warm` each had a target that could only be satisfied by adding content the source lacks (a first-person touch, a concrete claim or ask, an acknowledgment). Each target now applies only where the source already has the material, and the section opens with one line stating the guardrails bind voice targets. Wording ported back from the downstream resolution in wshobson/agents#645. Contributed by @mahinNadir (#133).
+
 ## [3.25.0] — 2026-08-12
 
 ### Added
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:18-slim
+
+WORKDIR /app
+
+COPY . .
+
+CMD ["npm", "test"]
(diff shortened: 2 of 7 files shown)


conorbronsdon:
Thanks for the contribution! Minor fixes coming shortly

conorbronsdon:
Thanks @mahinNadir — the wording matches what #100 prescribed, and the adversarial pass found no contradiction with the Never-inject section. I pushed a fix-up commit to your branch before merging: dropped the stray Dockerfile, gave the new guardrail line its own paragraph, fixed a comma splice in the professional profile, bumped plugin.json, regenerated the bundled plugin skill and cursor rule, and added the 3.25.1 changelog entry crediting you. Merging now.

PR #138, opened 2026-08-28
Add seven rhetorical-tic categories from Simon Willison's LLM cliché highlighter

t3chnaztea:
Adapts the pattern catalog from Simon Willison's [LLM cliché highlighter](https://tools.simonwillison.net/llm-cliche-highlighter) ([source](https://github.com/simonw/tools), Apache-2.0; regexes rewritten and tightened rather than copied, with per-rule credit inline). These are voice-of-the-essayist tics — the model performing insight rather than delivering it — and they complement the existing catalog: I checked each of the tool's 36 patterns against current main and dropped the ones you already cover (infomercial hooks, narrated candor, staccato drama, AI vocabulary, the Wikipedia-derived set).

**Three categories ship with detector types + fixtures (both directions):**

- `performed-insight` — "sit with that", "that's not nothing", "you already know the answer", "the punchline is", "worth naming", "don't take my word for it", sentence-initial "Turns out", "is the whole point / entire business model", "that's the part nobody mentions", "the only metric that matters", "is dead; long live", "that's why it mattered". Curated complements keep "the whole family" and "sit with him" clean. Weight 3 (density classifies, like tier2).
- `negation-chain` — "No fluff, no filler, no jargon" chains, stacked "didn't …" clauses, and the "Don't call it X. Call it Y." backref form. A stop-list keeps idiomatic pairs ("no more, no less", "no matter how") from firing, per the precision-over-recall rule. Weight 5 (strong single-hit structural tell).
- `dev-blog-boilerplate` — "batteries included", "it just works", "zero config", "sane defaults", "fits in your head". Weight 3.

**Four land as skill-only judgment rules,** with the reasons recorded in CATEGORIES.md group C: stacked rhetorical questions (interviews/FAQs stack questions legitimately), same-opener sentence runs (whether anaphora is earned is the whole judgment), stranded auxiliary contrast (one instance is legitimate style; only density distinguishes voice from tic), and colon into a triple (three-item lists are often simply true — Willison's own tool ships it off-by-default for docs).

Contracts updated per CONTRIBUTING: README bumped to 69 pattern categories and 51 engine types, CLAUDE.md copy matched, CATEGORIES.md rows + group C entries added, CHANGELOG 3.28.0, SKILL.md version bumped. `npm test` and `scripts/check-pattern-count.sh` pass.

Happy to split this into per-category PRs if that reviews easier.

https://claude.ai/code/session_01VNwGk41QYrtGFK7Utz1XfB


conorbronsdon:
Thanks @t3chnaztea for the contribution! Taking a look

conorbronsdon (changes requested):
## Review verdict: request changes

I reviewed this PR at exact commit `51cadbcda05c39d204a2958cf54e81df99879572`. The overall design is coherent: the seven additions split sensibly into three deterministic detector types and four judgment-only rules, the human-facing count moves from 62 to 69, the engine count moves from 48 to 51, and the category map agrees with `TYPE_LABELS`. The new happy-path and must-not-fire fixtures also pass.

There are still release-blocking synchronization omissions and several reproducible precision problems in the new regexes. Given this repository's explicit precision-over-recall policy, I don't think the detector findings should ship in their current form.

### 1. Blocking: the 3.28.0 release artifacts are not synchronized

The PR moves the root `SKILL.md` and top changelog entry to `3.28.0`, but four release/distribution surfaces remain at `3.27.0`:

- `package.json:3`
- `plugins/avoid-ai-writing/.claude-plugin/plugin.json:4`
- `plugins/avoid-ai-writing/skills/avoid-ai-writing/SKILL.md:4`
- `cursor-rules/avoid-ai-writing.mdc:2`

This is already a real CI failure, not a hypothetical one. `plugin-skill-sync / check` fails with:

```text
version mismatch: SKILL.md=3.28.0 plugin.json=3.27.0
Update plugin.json "version" to match SKILL.md frontmatter.
```

After the manifest is bumped, the workflow will regenerate the bundled skill and Cursor port and fail if those generated copies are not committed. Both copies currently still contain the 3.27.0 body and omit all seven new sections.

There is a second release failure waiting behind that check: `.github/workflows/release.yml` explicitly refuses npm publication when `package.json` disagrees with the top changelog version. On merge, the changelog push can create GitHub release/tag `v3.28.0`, but `publish-npm` will fail because `package.json` remains `3.27.0`. Fixing it in a later commit would then collide with the release workflow's exact-tag guard, potentially stranding the detector changes from npm under 3.28.0.

Required fix:

1. Bump `package.json` and the plugin manifest to `3.28.0`.
2. Run `bash scripts/sync-plugin-skill.sh && bash scripts/sync-cursor-rules.sh`.
3. Commit the generated skill and Cursor-rule changes.
4. Re-run the affected review/checks at the new SHA.

### 2. High: `negation-chain` flags ordinary narration and technical inventories

`detector/patterns.js:700` intentionally allows a period between two `did not` clauses and up to 25 intervening characters. That catches the documented clipped-rhythm case, but it also catches ordinary sequential narration:

```text
I did not sleep well last night. I did not eat breakfast either, so I left home early.
```

Verified output:

```text
type: negation-chain
severity: high
text: did not sleep well last night. I did not eat
```

The `no ...` branch at `patterns.js:694-699` similarly flags a normal technical contract:

```text
The endpoint takes no arguments, no headers, and no body when called in health-check mode.
```

Verified output:

```text
type: negation-chain
severity: high
text: no arguments, no headers, and no body when
```

These are especially costly false positives because this type is P1/high severity. The existing must-not-fire fixture tests only stop-list idioms (`no matter`, `no more, no less`) and never probes a near-miss with the same grammatical shape as the intended match.

Please narrow the deterministic subset or lower its severity, then add these ordinary narration and technical-inventory shapes as must-not-fire fixtures. Simply shortening the 25-character gap is unlikely to be enough: `I didn't sleep. I didn't eat.` is still ordinary prose despite being short.

### 3. Medium: two `performed-insight` branches overmatch ordinary prose or exceed the documented rule

I reproduced both of these at HEAD:

```text
Please sit with that decision overnight before you call the attorney tomorrow morning.
```

Flags `sit with that` at `patterns.js:672`. Here `that` is a determiner modifying `decision`, not the standalone reflective phrase documented as the tell. The expression needs a right-side guard so ordinary `sit with that <noun>` constructions stay clean.

```text
This is the only doctor I trust with a procedure this complicated and risky.
```

Flags `the only doctor I trust` at `patterns.js:682`. This is common literal English, and the new `SKILL.md` section documents “the only metric that matters,” not the broader “only X I/you/we/they trust” family. Either drop/narrow this branch or document why the broader family is precise enough to include.

The current negative fixture (`sat with him`, `the whole family`, `naming names`) does not exercise any of these near-boundary cases. These should be covered with explicit must-not-fire assertions after narrowing.

On reconsideration, I am **not** treating the broader `that's the part ...` branch at `patterns.js:681` as a defect. Even a concrete sentence such as “That is the part that keeps breaking” carries a mild stage-managed-emphasis shape, and `performed-insight` is appropriately weighted as a medium/3-point signal rather than decisive evidence on one hit. That makes the broader match defensible; the two cases above remain different because they capture ordinary grammatical continuations or an undocumented, commonplace literal construction.

### 4. Medium: the `Turns out` regex consumes the previous sentence boundary

At `detector/patterns.js:679`:

```js
/(?:^|[.!?]\s+|\n)Turns\s+out\b/g
```

For:

```text
A stable introduction goes here. Turns out the service was already running.
```

the detector reports:

```text
text: . Turns out
index: 31
```

The punctuation and whitespace belong to the preceding sentence, but become part of the issue text/index. That creates two downstream problems:

- sentence-region mapping can associate the issue start with the preceding sentence;
- deduplication keys on issue text, so the same phrase gets different keys depending on whether its boundary was a period or newline.

Please make the boundary non-consuming (for example, with the lookbehind idiom already used elsewhere in this file) and assert both the issue text and index in a regression fixture.

### 5. Low: `it just works` matches the different phrase `it just works out`

`detector/patterns.js:709` uses:

```js
/\bit\s+just\s+works\b/gi
```

That flags:

```text
It just works out to two queries after the optimizer merges identical branches in production.
```

“Works out to” is a phrasal verb meaning “evaluates/results in,” not the dev-marketing slogan “it just works.” A negative lookahead for continuations such as `out` would preserve the intended match without this false positive. Please add the reproduction as a must-not-fire fixture.

## Cross-model review and local verification

Required independent reviewers completed against this exact SHA:

- Claude: canonical model `claude-opus-5`, reference/completeness lens.
- Hermes/OpenRouter: exact model `thinkingmachines/inkling:free`, adversarial precision lens. OpenRouter's live catalog reported zero prompt and completion pricing when the review ran.

I treated both reports as unverified input and checked their claims against the files and runtime behavior. Notable discarded claims:

- Hermes claimed `[a-z]` in the negation regex rejects capitalized language names. It does not: the regex has the `i` flag, and `No Python, no Java` matches.
- Hermes treated lowercase `turns out` as a required match, but this PR's own catalog explicitly scopes that item to sentence-initial “Turns out”; I did not count the lowercase omission as a defect.
- Hermes's `feels worth naming` agreement concern did not demonstrate a regex bug.
- Opus predicted the changelog could exceed the self-scan budget. The actual check passed at 31/40, so I discarded that finding.
- I did not treat every phrase omitted from Simon Willison's broader implementation as a bug: this PR explicitly adapts and narrows the source inventory. I only retained divergences that conflict with this PR's own documented catalog or produce a concrete false positive.

Validation performed at `51cadbcda05c39d204a2958cf54e81df99879572`:

- `npm test`: all detector fixtures, the 51-type category contract, preservation validator, corpus helpers, and style checks passed.
- `npm run self-scan:check`: passed; all eight scanned documents remained within budget, including `CHANGELOG.md` at 31/40.
- Remote CI: detector test passed; plugin-skill-sync failed on the verified 3.28.0/3.27.0 manifest mismatch.
- PR head was rechecked immediately before this review and remained on the reviewed SHA.
- Review worktree remained clean; no code changes were made during review.

The direction is good, and the count/category wiring is solid. The release synchronization is mechanical. The important judgment call is the negation detector: a structural shape that appears frequently in legitimate narration and technical documentation should not be a high-severity signal unless the deterministic matcher can distinguish the rhetorical construction with acceptable precision.

t3chnaztea:
Thanks for the fast and thorough review — every reproduction checked out. All five findings addressed in d8b4d74:

1. **Release sync**: package.json, plugin.json bumped to 3.28.0; `sync-plugin-skill.sh` and `sync-cursor-rules.sh` run and the regenerated copies committed. plugin-skill-sync passes locally.
2. **negation-chain**: narrowed rather than downweighted. The `no …` chain now must open its sentence, so "takes no arguments, no headers, and no body" stays clean; the `did not` chain is comma-joined with the subject elided, so "I did not sleep well. I did not eat breakfast" stays clean. Both reproductions are must-not-fire fixtures, and SKILL.md documents the carve-outs (everything outside the narrow deterministic form is a judgment call).
3. **performed-insight**: `sit with (that|this)` now requires a following boundary or "for a moment"-family tail, so the determiner reading ("sit with that decision") stays clean. The `only X I trust` branch is dropped entirely — you're right that SKILL.md never documented it.
4. **Turns out**: switched to a lookbehind; fixture asserts both text (`Turns out`) and index.
5. **it just works**: negative lookahead for `out`; reproduction added to the boilerplate must-not-fire fixture.

`npm test` and `npm run self-scan:check` pass at the new head. On Qodo's quoted-cliché finding: the new families behave like the existing phrase families (no inline-quote masking anywhere in the engine), so I left that as-is rather than special-casing three types — happy to take it on as a follow-up if you want quote masking generally.

conorbronsdon:
Thanks! Taking a look at the new state - appreciate the speedy turnaround

conorbronsdon (approved):
Approved at `6acf9a1da1e0948750c2fc4c7a099f6b52279d23`.

The full test suite, self-scan, generated-copy checks, and pattern-count checks pass. I also reran the two independent reviews against this exact SHA: Claude Opus 5 and Inkling through Hermes found no blockers.

I made the small precision and documentation cleanups on the branch, including the changelog conflict resolution. The remaining edge cases are judgment calls rather than merge blockers, and I filed #140 so we can measure whether these carve-outs cost too much recall as the corpus grows.

Please do take the general inline quote masking as a separate follow-on PR. Reusing the current offset-preserving masker looks like the right direction. The tests should cover closed straight and curly quotes, apostrophes, escaped or unclosed quotes, an equivalent phrase outside quotes that must still fire, and preservation of source indices.

Thanks for offering to carry that forward.

conorbronsdon:
@t3chnaztea thanks again for contributing! Feel free to tag me in for review on the follow up if you take it

Issue #85, opened 2026-08-02
Human corpus has no docs, social, email, or conversational register — blocks every published FP number


conorbronsdon:
Split out of conorbronsdon/personal-context#119, which is now closed. Documented in `corpus/README.md` since the corpus landed, but never tracked as work.

## The gap

`corpus/manifest.json` covers, on the human leg:

| Register | Entries |
|---|---:|
| `blog` | 26 |
| `essay-literary` | 4 |
| `academic` | 3 |
| `mixed` | 2 |
| `technical-blog` | 1 |

Zero entries for `docs`, `social`, `email`, and `conversational` — four of the registers the tolerance matrix defines and the tool is pointed at every day. LinkedIn posts and release notes are the actual use case and neither is represented.

## Why this blocks more than it looks like

`corpus/README.md` states the gate and then fails it:

> no number from this corpus goes into the README, a release note, or a social post until each claim cell has n >= 100, covers more than one register that people actually write in today, and carries a confidence interval. The current run satisfies the interval and the n, and **fails the register test outright.**

So the measured 0.0% FP rate cannot be published. Patina's spread runs 4.0% on chat to 34.0% on technical how-to inside one language, which is exactly the shape a blog-only corpus hides. A single number from this corpus would be a blog number wearing a general-purpose label.

Downstream: #84 (cross-tool benchmark) is blocked on the same gap, since a cross-tool comparison on one register settles nothing.

## What would close it

Human-authored text, per register, with provenance someone will attest to, and n >= 100 units per cell.

- **`docs`** — pre-2023 technical documentation with a clear author and a permissive license. Plenty exists in open source, and the hash-only manifest means it does not have to be redistributable.
- **`social`** — hardest. Pre-2023 LinkedIn and X posts by the maintainer are the obvious seed, if exportable. The Wayback approach that recovered the pre-2023 blog posts is worth trying and may not work on these platforms.
- **`email`** — newsletter archives are the tractable form. Individual correspondence is not publishable in any form, hash-only included, and should not be attempted.
- **`conversational`** — least clear. Forum and mailing-list archives are pre-LLM, publicly licensed, and genuinely conversational.

Two intakes already exist: `node scripts/corpus.js add-local` for text that stays on disk, and the [false-positive report form](https://github.com/conorbronsdon/avoid-ai-writing/issues/new?template=false_positive.yml), which captures register precisely so reports accumulate into exactly this.

## Constraint

[...]

PR #141, opened 2026-08-29
docs: add seed coverage for missing corpus registers

UniversePeak:
## Summary

Adds auditable hash-only seed entries for the two corpus registers that the maintainer marked open to contributors:

- RFC 8259 as a pre-LLM `docs` source
- A 1995 W3C `www-talk` message as a pre-LLM `conversational` source

The corpus README now records that both registers remain under-sampled and that `social` and `email` remain empty because they require maintainer-provided or otherwise appropriate sources. This deliberately does not claim that either seed is sufficient to publish a false-positive rate.

The HTML extractor keeps authored semantic elements intact; the W3C archive entry uses its `main` container to exclude outer page chrome, while scripts and styles remain excluded. Focused regression fixtures cover both behaviors.

## Validation

- `node scripts/corpus.js verify` (new entries verify; pre-existing unavailable caches are reported as missing)
- `node scripts/corpus.test.js`
- `npm test`

Contributes to #85

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project are documented here.
 
 ---
 
-## [Unreleased]
+## [Unreleased] — 2026-08-29
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ All notable changes to this project are documented here.
 - **README adds a restrained related-work block.** Links to Conor's public
   builds, Chain of Thought, and `repo-audit` now sit after the core product and
   usage documentation.
+- **Corpus manifest documents the register gap with two auditable seed entries.**
+  RFC 8259 provides a pre-LLM `docs` source and a 1995 W3C mailing-list message
+  provides a pre-LLM `conversational` source. The corpus README records that
+  both registers remain under-sampled and that `social` and `email` still have
+  no entries; the additions do not authorize publishing a rate.
 
 ---
 
--- a/corpus/README.md
+++ b/corpus/README.md
@@ -44,7 +44,9 @@ buckets are what let that assertion be checked instead of assumed.
 
 ## Current contents
 
-Two sources, chosen for different reasons.
+Four source groups, chosen for different reasons. The last two are deliberately
+small seed entries: they make the missing-register state visible in the manifest
+without pretending that one document is enough to support a rate.
 
 **Nine public-domain works, 1788 to 1907**, sliced to 6,000 words each. Their
 provenance is beyond argument: nothing written in 1859 was machine-generated.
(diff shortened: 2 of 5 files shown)


conorbronsdon:
Thanks for the contribution @UniversePeak! Taking a look

conorbronsdon (changes requested):
Reviewed commit: `8f3a11b8ca9c031132efaad94056dbd6a608e346`

Requesting changes for four corpus-integrity and workflow issues:

1. **Do not auto-close #85.** The issue says partial contributions are useful but `n >= 100` per register is required before a cell supports a publishable claim. This PR adds one `docs` and one `conversational` source, leaves both under-sampled, and leaves `social` and `email` empty. Please replace `Fixes #85` with `Contributes to #85` or equivalent so the remaining work stays tracked.

2. **Exclude RFC front matter from the measured `docs` sample.** The `slice.after: "Abstract"` boundary retains the Status section, Copyright Notice, and table of contents before the actual technical prose. Direct source verification found roughly 879 of the recorded 3,967 words in that front matter; three of those boilerplate paragraphs enter paragraph-level measurement, and all of it enters document-level measurement. `slice.after: "1.  Introduction"` is a unique boundary in the official text; use that or another explicit front-matter stripper, then regenerate `sha256` and `words`.

3. **Do not globally strip semantic `header`, `footer`, and `nav` elements to support this source.** The replacement in `htmlToText()` deletes those elements inside every selected container, including authored `<article><header>` content. A 60-word control header is removed completely. It is also unnecessary for this W3C entry: the old and new extractors produce the same source hash because `selector: "main"` already excludes the outer page chrome. Remove the global change or make chrome stripping an explicit per-source option.

4. **Record an actual license or rights status for the W3C message.** `"W3C public mailing-list archive"` describes where the message is hosted, not the applicable license. W3C's mailing-list policy establishes public archival consent, but the source page does not identify a license for this individual message. Cite the applicable license if one exists, or record an accurate author-retained/not-redistributed or unverified rights status consistent with the hash-only corpus policy.

Verification performed at this SHA:

- `npm.cmd test` passed all suites; the focused corpus suite passed 21 tests.
- Both live sources reproduced their recorded hashes and word counts exactly.
- GitHub's `test` and `check` jobs are green.
- Claude Sonnet 5 and Hermes `thinkingmachines/inkling:free` completed independent reviews; their claims were checked against the files and live public sources.

I did not accept Qodo's suggestion that this PR itself must populate all four registers to 100 units; partial contributions are explicitly allowed. The problem is only the `Fixes #85` auto-close language.

UniversePeak:
Addressed the maintainer review in commit `a3c9e56`:

- Changed `Fixes #85` to `Contributes to #85` so the remaining register work stays tracked.
- Changed the RFC 8259 slice boundary to `1.  Introduction`; regenerated the measured cache and recorded SHA-256 `b593bd1763d8f3e506356b1d4c83ec0faf57a133e1a31684a4c693ec27694573` over 3,601 words.
- Removed global stripping of semantic `header`/`nav`/`footer`; added regression coverage for authored semantic content and selector-scoped page chrome.
- Replaced the W3C archive label with an unverified author-retained/not-redistributed rights status.
- Dated the unreleased changelog heading without inventing a release version.

Local verification: `npm.cmd test` passed; `node scripts/corpus.js verify` reports both new entries `ok` (pre-existing unavailable caches remain reported as missing). The PR checks are green. The `social`/`email` zero-coverage state remains intentional and documented: this contribution is partial, while #85 stays open until the full register requirements are met.

conorbronsdon:
Thank you @UniversePeak! Deeply appreciate the contribution.

conorbronsdon (approved):

PR #145, opened 2026-09-04
Harden OpenAI plugin validation for deferred port gaps

### LINKED ISSUE #143: Follow-ups deferred from the ChatGPT/Codex port review (#135)
Deferred findings from the #135 review rounds (Codex, Qodo, and two open models), none blocking the port itself. Grouped by where the fix belongs.

## Preservation validator (`detector/validate.js`), pre-existing on main

Reproduced against `main`'s own copy, so they are not the port's bugs:

1. A table cell changed in a table without outer pipes passes (`{"ok":true}`).
2. Content after a `~~~` line inside a backtick-fenced block changed passes.
3. The documented `utm_source` strip with a second query parameter reports `url-missing` instead of passing (query-preserving cleanup).
4. Indented (four-space) code block changes are not compared exactly.

Each needs a fixture pair in `detector/validate.test.js` and a targeted fix; the bundled copies under `skills/` regenerate from the source via `scripts/sync-plugin-skill.sh`.

## Canonical SKILL.md content, pre-existing on main

5. The word-replacement row `beacon | (rewrite entirely)` gives no concrete alternative (the table rule says every entry lists specific alternatives).
6. The engine scans Tier 1 tokens inside ordinary quotations and explicitly illustrative examples; the self-reference escape hatch is prose-only.
7. `--context technical` still flags the documented technical-blog exceptions (`robust`, `ecosystem`); the exception list is prose-only.

Items 6 and 7 are detector design decisions, not one-line fixes: masking quoted examples and honoring the technical exception list in `detector/patterns.js` changes recall for every consumer.

## OpenAI plugin validator strictness (`scripts/validate-openai-plugin.py`)

8. A scalar `policy:` value in `agents/openai.yaml` passes; require a mapping.
9. An SVG whose root element is not `<svg>` passes; check the root tag.
10. Malformed `agents/openai.yaml` that is not the expected key/value shape passes; fail on unparseable lines.
11. `skills/avoid-ai-writing-router/references/routing-matrix.md` restates `references/skill-graph.json` in prose and nothing compares them; either generate the matrix from the graph or add a drift check.

## Design question

12. `skills/voice-preserving-rewriter/SKILL.md` carries its own short `## Output` section. The repo's single-file rule was written for the canonical skill and its generated copies; the port's specialized skills route to the canonical copy but this one restates an output contract. Decide whether it should point at the canonical format instead.

## Portal-side, not code

13. Verified developer identity on the OpenAI portal and a test of the exact final archive; `submission/submission-pack.json` correctly stays `not_ready` until then.

UniversePeak:
## Summary

Addresses the implementation-owned follow-ups in #143 from the ChatGPT/Codex port review:

- Reject scalar `policy:` values and malformed `agents/openai.yaml` mapping/list lines instead of allowing invalid metadata through.
- Require SVG assets to have an actual `<svg>` root element.
- Make the human-readable routing matrix generated-and-verified against `skill-graph.json` with a SHA-256 graph digest and exact edge inventory, preventing silent drift.

## Validation

- `python scripts/validate-openai-plugin.test.py`
- `python -m py_compile scripts/validate-openai-plugin.py`
- `python skills/avoid-ai-writing-router/scripts/validate_connections.py .`
- `python scripts/validate-openai-plugin.py .`
- `npm test`
- `git diff --check`

Closes #143.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project are documented here.
 
 ---
 
-## [Unreleased] — 2026-08-29
+## [Unreleased] — 2026-09-04
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project are documented here.
 
 ### Changed
 
+- **Plugin validation now fails closed on deferred port-integrity gaps.** `agents/openai.yaml` rejects scalar policies and malformed mapping/list lines, SVG assets must have an actual `<svg>` root, and the bundled routing matrix carries a checked graph digest plus generated edge inventory so it cannot silently drift from `skill-graph.json`.
+
 - **README adds a restrained related-work block.** Links to Conor's public
   builds, Chain of Thought, and `repo-audit` now sit after the core product and
   usage documentation.
--- a/scripts/validate-openai-plugin.py
+++ b/scripts/validate-openai-plugin.py
@@ -226,12 +226,16 @@ def validate_yaml_shape(path: Path, text: str, errors):
                 current_section = None
                 current_nested = None
                 continue
-            key = match.group(1)
+            key, value = match.groups()
+            value = value or ""
             if key in top_keys:
                 errors.append(f"{path}: duplicate top-level key: {key}")
             top_keys.add(key)
             if key not in AGENT_METADATA_KEYS:
                 errors.append(f"{path}: unknown top-level key: {key}")
+                current_section = None
+            elif value.strip() and not value.strip().startswith("#"):
+                errors.append(f"{path}: malformed YAML line {number}: {key} must be a mapping")
                 current_section = None
             else:
                 current_section = key
--- a/scripts/validate-openai-plugin.py
+++ b/scripts/validate-openai-plugin.py
@@ -129,11 +129,12 @@ def validate_agent_metadata(path: Path, errors):
                 stripped = line.strip()
                 if not stripped or stripped.startswith("#"):
                     continue
-                match = re.fullmatch(r"-\s*([^#]+?)(?:\s+#.*)?", stripped)
-                if not match:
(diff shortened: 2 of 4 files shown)


conorbronsdon:
Thanks @UniversePeak I'm taking a look

conorbronsdon (changes requested):
Requesting changes based on two reproduced fail-open cases at c4abb87d6940914b06fe00baedbb9d0ad97c3721.

1. Malformed YAML without whitespace after a mapping colon passes validation. For example, adding `products:[CHAT]` under `policy` makes `validate_agent_metadata` return no errors, while PyYAML 6.0.3 raises `ScannerError: could not find expected colon`. The new mapping regexes use `:\s*`, so they accept zero whitespace. Please require whitespace or end-of-line after mapping colons in both the shape and policy parsers, and add a regression test.

2. Required keys are still detected with raw substring searches. A commented `# short_description: missing` satisfies the check even though parsed YAML has no `short_description`; likewise, `default_prompt: "policy: allow_implicit_invocation:"` satisfies both policy tokens while the parsed document has no `policy` mapping. Both fixtures return no validation errors. Please assert required top-level and nested keys from the parsed shape state instead of searching raw text, with must-fail regression tests.

The full existing Python and npm suites pass, which is why targeted mutations are needed here. OpenCode independently identified the first case; I reproduced both locally against this exact commit.

UniversePeak:
Addressed the two requested changes in commit e84755e317251e19120fcdcc12dd60c84bcd0f6c:

- Mapping validation now requires whitespace or end-of-line after `:`, with regression coverage for `products:[CHAT]`.
- Required `interface`/`policy` keys are checked from the parsed shape state, so comments and quoted values cannot satisfy them; regression coverage covers both cases.

The existing fail-closed YAML, SVG, routing-matrix, and malformed-edge checks were retained. Focused Python validation, `py_compile`, router connection validation, `npm test`, and `git diff --check` pass. Please take another look when convenient.

conorbronsdon (approved):

PR #178, opened 2026-09-12
fix: define evidence policy for stock reaction framing

### LINKED ISSUE #82: Test the emotional-flatline rule — StoryScope measures the opposite direction
Split out of conorbronsdon/personal-context#119 (Tier 3), which is now closed so this has one home.

## The claim under test

`SKILL.md` lines 443–448, "Emotional flatline," treats a named emotion as the AI tell:

> AI claims emotions as a structural crutch without conveying them through the writing: "What surprised me most," "I was fascinated to discover," "What struck me was"…
> it's tell-don't-show: if the thing is genuinely surprising, the reader should feel that from the content, not from the writer announcing it.

## The measurement that points the other way

StoryScope (Russell et al. 2026, arXiv:2604.03136) reports the opposite split on human versus machine narrative:

| Behaviour | AI | Human |
|---|---:|---:|
| Performs emotion through the body (racing heart, tight chest) | 81% | 38% |
| Names the feeling outright | 8% | 29% |

On that data, "show, don't tell" is the machine habit and naming the feeling is the human one. The rule as written pushes authors toward the AI-side behaviour.

## Why this one is worth testing first

It is the most falsifiable claim in the rule set. The rule prescribes a specific direction, an external study measures that exact direction, and the two disagree. Most entries in the catalog are style advice where "correct" is contestable. This one has a testable premise underneath it.

## What resolving it takes

1. Decide whether StoryScope's finding transfers. Its corpus is narrative fiction; this skill targets blog posts, LinkedIn, release notes, and docs. The transfer is an assumption, not a given, and it is the first thing to check.
2. Measure on this repo's own corpus. The machine leg (RAID, HC3) and the human leg (verified pre-2023 posts) already exist. Firing rate of the `emotional-flatline` patterns on each side gives a lift number comparable to the table in #71.
3. Act on the result:
   - lift well above 1 — the rule stands, and it now has evidence rather than intuition
   - lift near 1 — reframe it as style advice rather than an authorship signal, the treatment #73 proposes for `em-dash`
   - lift below 1 — the rule is inverted, same as `em-dash`, and the phrasing has to change

## Caveat carried from #72

Every machine unit in the corpus predates 2025, so a measurement here is an upper bound on current models. That is a reason to record the caveat with the number, not a reason to skip the measurement: an inverted rule is worth knowing about even on old data, and `em-dash` was found exactly this way.

### LINKED ISSUE #179: Prevent heading attachment from dropping an oversized measurement unit
﻿## Problem

PR #178 teaches paragraph-level corpus measurement to attach a short heading block to the paragraph that follows it. The code advances past the paragraph before checking the combined word count.

If the following paragraph is valid by itself at the 400-word ceiling, adding even a one-word heading makes the combined unit too long. The combined unit is rejected and the consumed paragraph is never measured.

Reproduction on PR #178 head `8f5df8bc36014c08c700013b8b112ef4d2bceae2`:

```json
{"standalone":1,"afterHeading":0}
```

## Expected behavior

Check the combined size before consuming the following block. If the combined unit exceeds `MAX_WORDS`, retain the following paragraph as its own measurement unit.

Add a regression proving that a 400-word paragraph remains measurable when preceded by a short heading.

Blocks #178. Follow-up to #82.

conorbronsdon:
﻿PR #177 found a real gap: the `emotional-flatline` rule had no corpus evidence behind its authorship weight, and its writing guidance could push authors toward theatrical body language. Its measurement could not establish a direction, however, and the StoryScope paper studies a different construct in narrative fiction.

This maintainer follow-up preserves Francisco Quinteros's work in an author-attributed commit, then tightens the method and policy around it.

### What changes

- Rename the user-facing category to **Stock reaction framing** while keeping the `emotional-flatline` API type stable.
- Preserve specific reactions and flag only empty stock framing.
- Keep the finding visible but give it zero authorship weight under an explicit precision-first policy: an unvalidated style signal does not move the score.
- Keep zero-weight findings out of AI sentence highlights.
- Record the corpus result as inconclusive and explain why StoryScope cannot determine this category's direction.
- Join hard-wrapped corpus prose while preserving semantic paragraph and heading boundaries, so line-anchored patterns remain observable without treating source-format wraps as headings.
- Add symmetric controls for headings, running prose, and hard wraps in both document and paragraph preprocessing.
- Publish the behavior as version 3.35.0 and regenerate every distributed copy.

### Evidence

The corrected measurement still observes no examples of the category in either corpus class. The preprocessing defect therefore did not create the null result. It remains an evidence gap, not evidence of equivalence or neutrality.

Validation:

- `npm test`
- `npm run self-scan:check`
- `bash scripts/sync-plugin-skill.sh`
- `bash scripts/sync-cursor-rules.sh`
- corrected document and paragraph corpus measurements

Draft while the replacement exact-SHA review is completed.

Builds on #177. Closes #82. Closes #179.


conorbronsdon:
﻿The exact-SHA review found one remaining measurement blocker, now tracked in https://github.com/conorbronsdon/avoid-ai-writing/issues/179.

A short heading currently consumes the following paragraph before the combined MAX_WORDS check. The reproduced case keeps a standalone 400-word paragraph but drops it when preceded by Context:. I'm leaving this PR in draft until that case is fixed and re-reviewed.

JavierQuinan (commented):
I reviewed the maintainer follow-up with focus on the evidence boundary, preprocessing semantics, scoring/highlight consistency, release/version synchronization, and generated-copy checks.

The policy split is now much clearer: the corpus result is explicitly inconclusive, StoryScope is treated as a different construct/register, and the zero-weight behavior is stated as a precision-first policy rather than inferred from 0/0 evidence. Version 3.35.0 is synchronized across the canonical skill, package, Claude/Codex manifests, and submission metadata; detector, plugin-skill-sync, and SSOT are green.

The remaining blocker I could reproduce is #179. I opened #180 against this branch with a minimal fix and two boundary regressions. I would wait for #180 to land and for the maintainer's replacement exact-SHA eval before considering #178 ready. I did not find another blocking issue in the canonical surfaces I reviewed.

conorbronsdon:
Codex review, with GPT-5.6 Sol reviewers, at Conor's request. Reviewed head a37404a9545e8b766740737520f76725cf471d65.

The clearer Stock reaction framing name, stable emotional-flatline API type, isolated regression, and distinction between StoryScope's fiction construct and this regex are useful. #287 carries those edits into the first claimant's #177 history, with explicit attribution. Thank you also for preserving Francisco's work and his #180 boundary fix here.

The preprocessing portion needs larger revisions and remains open. I reproduced:

- `Interesting part of the project:\n` plus a 400-word body has an emotional-flatline hit in raw text that normalizeUnit removes in document mode.
- Five blockquoted `What surprised me most...` lines are clean as raw quoted text but become a live emotional-flatline finding after normalization.
- Five short noun-phrase bullets have a bullet-np-list finding that disappears when normalization flattens them.

Keep #180's correct lookahead/single-newline boundary work in a focused preprocessing successor, add semantic-boundary regressions for these cases, and rerun the corpus measurement on that exact head. #287 does not include fp-measure changes or release-version work; #179 remains open. I do not recommend merging this PR as-is.

conorbronsdon:
Coordination update from Codex: #288 now tracks the focused preprocessing repair, preservation of #180's contribution, exact-commit corpus comparisons, and independent GPT/Claude/additional-model-family reviews. Please claim an implementation or review role there before starting so the sessions can work together. This session is currently setting up review clients; it has not claimed repair implementation.

conorbronsdon:
Codex follow-up: three independent OpenCode model reviews of this head (`a37404a`) are complete, followed by local execution checks. The [consolidated review on #288](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5655821406) records the exact models/sessions, verified reproductions, and model claims that did not hold up. Conor is also arranging a separate Claude review.

The 15 existing preprocessing tests pass, but document-mode heading loss and the lowercase-starting body gap remain reproducible. Quote/list flattening is also confirmed in both main and this head. Recommendation remains a focused successor from current main, preserving #180's correct boundary behavior and Francisco's credit, followed by fixed-corpus measurements and review of the repaired commit. No successor or final merge approval is claimed by this review pass.

conorbronsdon:
The remaining preprocessing work now has a focused successor on current main: #290 at `f1dfd940559ab8cf3041ec6465ac79a0ad0b9e72`. It carries #180's boundary fix/credit, preserves document content and detector-visible structure, and adds verified-corpus decision accounting. All 18 suites, all three CI workflows, and final Sol/MiMo/Ling/Nemotron review checks are complete. [Exact results and reproducible evidence](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5656136957). Keeping this PR open until the separate Claude final-head review is recorded and #290 lands; then it can close as superseded.

conorbronsdon:
Codex maintainer closeout: the remaining preprocessing repair has landed in #290, merge commit `940a8812647ada70a97eb16b24e879dfa1512422`. The already-landed policy work remains in #177/#287, and #179 is now closed as fixed.

Thanks to @JavierQuinan for the #180 boundary fix and regression. Francisco's contribution is preserved in the implementation history and co-author trailers. Thanks also to the independent reviewers for reproducing the defects and checking the successor. [Claude's final review](https://github.com/conorbronsdon/avoid-ai-writing/pull/290#pullrequestreview-5192417046) cleared the final gate on `86ef5ab3` after reproducing the verified 38-source comparison byte-for-byte.

Closing this PR as superseded by the landed successor. Non-blocking follow-ups are tracked in #293; no remaining changes from this branch need to be merged.

PR #195, opened 2026-09-12
ci: reject release version drift before tagging

### LINKED ISSUE #191: Reject release version drift before creating a tag or GitHub release
The release workflow warns when package.json disagrees with the latest numeric changelog heading, then can create a tag and GitHub release for the mismatched commit. Only the later npm job makes that mismatch an error.

Verified by inspecting [.github/workflows/release.yml](https://github.com/conorbronsdon/avoid-ai-writing/blob/d9adf2c815b6cd929825f8076ec62dc1ba6ad82c/.github/workflows/release.yml) on main `d9adf2c815b6cd929825f8076ec62dc1ba6ad82c`: the initial version check warns (around lines 36–51), `gh release create` runs afterward, and the npm job rejects drift around lines 132–144. No live release was created during this review.

For example, a new 3.35.0 changelog heading with package.json still at 3.34.0 can create v3.35.0 and then fail npm publishing. Fixing package.json in a later commit leaves the tag on the earlier commit; the existing provenance guard correctly refuses to publish that later commit under the old tag. Prevent the partial release before it happens.

### Scope and acceptance criteria

- Reject missing, malformed, or mismatched package/changelog versions before any tag or release mutation.
- Keep version extraction and comparison consistent across the release and npm jobs; retain the existing commit/tag provenance and already-published checks.
- Cover matching versions, mismatches, missing or invalid versions/headings, and Unreleased-only documentation edits.
- Use read-only fixtures or mocked release commands to prove invalid input reaches no create/publish command; do not create test releases or publish test packages.
- Matching release preparation and already-published versions remain idempotent.
- Workflow syntax and `npm test` pass.

This is a focused release-engineering task for `help wanted`; it does not authorize moving existing tags or repairing a live release.

sharadvc:
## Summary

Fails the release job when `package.json` and the latest numeric `CHANGELOG.md` heading disagree, instead of only warning and still running `gh release create`.

- Adds `scripts/verify-release-versions.js` shared guard
- Wires it into `.github/workflows/release.yml` for release + publish-npm
- Adds fixture tests covering match, drift, and bad inputs

## Test plan

- [x] `npm test` (includes new verify-release-versions cases)

Fixes #191

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,20 +35,7 @@ jobs:
 
       - name: Parse latest version from CHANGELOG.md
         id: changelog
-        run: |
-          HEADING_LINE="$(grep -m1 -E '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md || true)"
-          if [ -z "$HEADING_LINE" ]; then
-            echo "::error::Could not find a '## [X.Y.Z]' heading in CHANGELOG.md"
-            exit 1
-          fi
-          VERSION="$(echo "$HEADING_LINE" | sed -E 's/^## \[([0-9]+\.[0-9]+\.[0-9]+)\].*/\1/')"
-          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
-          echo "Latest CHANGELOG.md version: $VERSION"
-
-          PKG_VERSION="$(node -p "require('./package.json').version" 2>/dev/null || echo unknown)"
-          if [ "$PKG_VERSION" != "$VERSION" ]; then
-            echo "::warning::package.json version ($PKG_VERSION) does not match the top CHANGELOG.md entry ($VERSION). This is pre-existing drift, not something this workflow fixes."
-          fi
+        run: node scripts/verify-release-versions.js --github-output "$GITHUB_OUTPUT"
 
       - name: Check if the release tag already exists
         id: tagcheck
@@ -96,8 +83,8 @@ jobs:
   # all. This job closes that gap. It is independently idempotent (skips when
   # the version is already on npm), so re-runs and already-tagged pushes are
   # safe, and it publishes package.json's version only after proving it matches
-  # the CHANGELOG heading the release job used — the drift this workflow
-  # already warns about must block a publish, not ride into the registry.
+  # the CHANGELOG heading the release job used — the same version guard as the
+  # release job must block a publish, not ride into the registry.
   publish-npm:
     needs: release
     runs-on: ubuntu-latest
@@ -130,18 +117,7 @@ jobs:
           '
 
       - name: Refuse to publish a version that disagrees with CHANGELOG.md
-        run: |
-          PKG_VERSION="$(node -p "require('./package.json').version")"
-          HEADING_LINE="$(grep -m1 -E '^## \[[0-9]+\.[0-9]+\.[0-9]+\]' CHANGELOG.md || true)"
-          if [ -z "$HEADING_LINE" ]; then
-            echo "::error::Could not find a '## [X.Y.Z]' heading in CHANGELOG.md"
-            exit 1
-          fi
-          CHANGELOG_VERSION="$(echo "$HEADING_LINE" | sed -E 's/^## \[([0-9]+\.[0-9]+\.[0-9]+)\].*/\1/')"
-          if [ "$PKG_VERSION" != "$CHANGELOG_VERSION" ]; then
-            echo "::error::package.json ($PKG_VERSION) != CHANGELOG.md ($CHANGELOG_VERSION) — fix the drift, then re-run. Publishing a mismatched version bakes the drift into the registry."
-            exit 1
-          fi
+        run: node scripts/verify-release-versions.js
 
       - name: Skip if this version is already on npm
         id: npmcheck
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "node": ">=18"
   },
   "scripts": {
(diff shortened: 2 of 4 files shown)


conorbronsdon (changes requested):
Thanks—putting a shared verifier before release mutations is the right shape, and the existing tests/full suite pass. There is one blocking parser case from #191: `readChangelogVersion()` skips malformed release headings until it finds any older numeric heading.

For example, a changelog with `## [Unreleased]`, then `## [3.35]`, then the prior `## [3.34.0]`, paired with package version `3.34.0`, returns success. That could release the old version even though the newest release heading is malformed; the acceptance criteria require malformed headings to fail before tag/release/publish mutations.

Please treat the first release heading after Unreleased as authoritative and validate it instead of skipping invalid headings. Add regression coverage for that fallback case and for a missing `package.json` version. It would also strengthen the workflow proof to assert that neither release creation nor npm publication is reached for invalid inputs. The workflow wiring and matching-version path otherwise look good.

conorbronsdon (changes requested):
## Claude's review

Reviewed independently against #191. The shape is right: one guard, shared by both jobs, ahead of every mutation. I confirmed the wiring preserves what #191 asked to keep — the guard runs before the tag check and before `gh release create`, `publish-npm` still carries the already-published skip and the tag-provenance refusal, and the branch merges cleanly into `c5a8b05` with `npm test` passing there. Run against the real repo the guard agrees on 3.34.0.

Requesting changes for the parser case @conorbronsdon already raised, plus four items I did not see covered.

### Confirmed: malformed headings are skipped, not rejected

I reproduced this independently. A changelog of `## [3.34]` followed by `## [3.33.0]`, with `package.json` at `3.33.0`, returns `ok: true` on version 3.33.0. `readChangelogVersion()` walks past any line the regex does not match, so a typo in the newest heading silently promotes an older release rather than failing. #191 asks for malformed versions to be rejected before any mutation, so I agree this blocks.

### New: `npm test` now requires bash and a POSIX PATH

The CLI isolation test spawns `bash` directly, writes a `#!/bin/sh` mock, and joins PATH with a literal colon. Every other test in the suite spawns `process.execPath` or `git`; this is the first to need a shell. Reproduced by running the file with bash off PATH:

```
actual: null, expected: 0, operator: 'strictEqual'
```

`spawnSync` returns no status when the binary is missing, so the assertion fails before anything else runs. `CONTRIBUTING.md` promises "No dependencies to install; Node 18+ only" and `engines` only pins Node, so this silently narrows who can run the suite. Model the sequencing in Node, and use `path.delimiter` for the PATH edit.

### New: the simulated release harness has a dead success path

Inside the bash block the script does `cd "$root"` and then requires `./verify-release-versions.js`, but the fixture root contains only `CHANGELOG.md` and `package.json`. That require can never resolve.

The test still discriminates today — I disabled the drift comparison and the mock `gh` ran, so the `ghLog` assertion caught the mutant. But it caught it for the right reason by luck: the success half of the simulation is unreachable, and the `status === 0` assertion passes through the early-exit branch rather than through a real release path. Anyone extending this to assert a tag name will be debugging a require that was never going to work. Point the require at `scriptPath`.

### New: an empty output path exits 0 instead of failing closed

`--github-output ""` with `GITHUB_OUTPUT` unset exits 0, prints the agreement line, and writes nothing. The shell step this replaces would have died on `>> ""`. In the workflow that leaves `steps.changelog.outputs.version` empty and `gh release create "v"` downstream. Treat a missing or empty path as an error.

### New: the most common trigger has no fixture

The workflow fires on any push to `main` touching `CHANGELOG.md`, so the everyday case is an Unreleased-only edit sitting above a prior release heading with `package.json` already matching it. `readChangelogVersion` covers the parse, but no `verifyReleaseVersions` fixture pins that end-to-end shape as passing — and #191 lists it explicitly. A `package.json` with no `version` field is also handled in code and untested.

### Nit, not a regression

`SEMVER_RE` accepts leading zeros, so matching `01.2.3` on both sides passes and tags `v01.2.3`. The shell it replaces used the same pattern, so nothing regresses; `(?:0|[1-9][0-9]*)` closes it cheaply. Separately, prerelease versions like `3.35.0-rc.1` now hard-fail the release job where they previously only warned. That reads intentional, but it is a behavior change worth one line in the script comment.

### Where I disagree with the automated review

Two of its findings ask for work you should not do. There is no changelog entry or version bump owed here. The current `CONTRIBUTING.md` exempts "internal refactoring or CI maintenance with no user-facing effect" and moves `SKILL.md`, `package.json`, and both plugin manifests at release preparation, not per change. This PR touches no shipped behavior, so leaving all four at 3.34.0 is correct.

---

conorbronsdon (changes requested):

Thanks, @sharadvc, for moving the release guard ahead of tagging and sharing it across both jobs. This still needs a substantive revision before merge. I reproduced the existing blocker: with Unreleased followed by malformed `## [3.35]`, then old `## [3.34.0]`, and package version 3.34.0, the guard returns success by skipping the malformed current heading. It must fail there rather than select an older release.

Please add that regression, a missing-package-version case, and an end-to-end Unreleased-only case. The test harness should use Node-portable sequencing instead of bash/POSIX PATH assumptions, and an empty `--github-output` must fail clearly. Reconcile package.json with the new test runner and preserve every suite when resolving the current conflict.

The existing CI is green but does not exercise the malformed-heading fallback. I am leaving this open for your revision; the core shared-guard approach remains useful.

conorbronsdon:
Thanks again, @sharadvc. I rescued this onto current main in #303, kept your original commit and authorship intact as 9b0cae84befeaee2a13d0f19648a076ab19493e5, and applied the requested parser, portability, and failure-path fixes in a separate maintainer commit.\n\n#303 is now merged, so I’m closing this superseded PR.

PR #196, opened 2026-09-12
perf: make HTML comment masking linear in comment count

### LINKED ISSUE #190: Make rendered-Markdown HTML comment masking linear
Inputs with many HTML comments spend increasing time in `maskRenderedMarkdown()` before ordinary word-count limits run. The loop calls `maskCommentCode()` for every comment; that helper joins and re-masks the complete document each time. With c comments over n characters, this prepass performs O(c × n) work.

Verified on main `d9adf2c815b6cd929825f8076ec62dc1ba6ad82c`; see [detector/patterns.js](https://github.com/conorbronsdon/avoid-ai-writing/blob/d9adf2c815b6cd929825f8076ec62dc1ba6ad82c/detector/patterns.js), `maskRenderedMarkdown()` around lines 931–958 and the later `MAX_WORDS` check. Comments add no counted words, so the word limit does not bound this preprocessing.

### Reproduction

```js
const d = require('./detector/patterns.js');
for (const n of [100, 500, 1000, 2000]) {
  const text = '<!-- x -->\n'.repeat(n) +
    'one two three four five six seven eight nine ten';
  const start = process.hrtime.bigint();
  d.analyzeText(text, { sourceMode: 'rendered-markdown' });
  console.log(n, Number(process.hrtime.bigint() - start) / 1e6);
}
```

Two independent local checks reproduced the growth. One measured about 20 ms for 100 comments, 202 ms for 500, and 782 ms for 1,000. These timings describe that environment; the repeated whole-document scan is the regression to prevent.

### Acceptance criteria

- Scan comment/code structure in a bounded single or small number of passes, without a full-document re-mask for each comment.
- Preserve real-comment masking, unclosed comments through EOF, and comment markers inside fenced, inline, or top-level indented code.
- Preserve the cases where code-looking delimiters inside an already masked comment must not change later comment recognition.
- Keep source offsets and `maskedHtmlComments` correct.
- Add semantic fixtures and a high-comment-count performance regression or deterministic control that catches repeated full-document rescanning without a fragile timing threshold.
- Regenerate detector mirrors; `npm test` and `npm run self-scan:check` pass.

This needs care around Markdown parsing and is intended for `help wanted`, rather than a first contribution.

sharadvc:
## Summary

`maskRenderedMarkdown()` no longer re-runs full code masking for every HTML comment (O(comments × document)). It masks code once, scans openings forward, and only refreshes the code mask when a removed comment contained backticks or list markers.

## Test plan

- [x] `npm test` (includes #190 fixture: 2000 comments, runtime bound)
- [x] `npm run self-scan:check`

Fixes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project are documented here.
 
 ### Fixed
 
-- Include every observed corpus register in `corpus.js list`; preserve the preferred accepted-register order and sort additional registers deterministically.
+- Make rendered-Markdown HTML comment masking linear: run `maskCode()` once per document, refresh code and list-aware scans only when a removed comment contained backticks or list markers, and scan for openings in a single forward pass (#190).
 - Fix three README link targets: the dead Cowork URL, the pattern-catalog pointer, and the
   voice-profile link that led to the triggering section.
 - Correct the `analyzeText()` result table in `detector/README.md`: the six score labels the
--- a/detector/patterns.js
+++ b/detector/patterns.js
@@ -996,6 +996,136 @@ const AIDetector = (() => {
     return { start: 0, end: lines[closingLine].end };
   }
 
+  // Mask HTML comments in source order while tracking the Markdown constructs
+  // that protect a literal `<!--`. A comment wins over code delimiters that
+  // occur inside it; a fence, code span, or top-level indented block that
+  // starts first wins over comment-looking text inside that code. Each source
+  // character participates in a bounded number of forward scans.
+  function maskHtmlCommentsOutsideCode(chars) {
+    const source = chars.join('');
+    const lines = source.split('\n');
+    let offset = 0;
+    let openFence = null;
+    let inIndentedBlock = false;
+    let previousBlank = true;
+    let listContext = false;
+    let maskedHtmlComments = 0;
+    const commentClosings = [];
+    let closingCursor = 0;
+
+    for (let i = 0; i <= source.length - 3; i += 1) {
+      if (source[i] === '-' && source[i + 1] === '-' && source[i + 2] === '>') {
+        commentClosings.push(i);
+      }
+    }
+
+    const backtickRuns = (line) => {
+      const runs = [];
+      for (let i = 0; i < line.length;) {
+        if (line[i] !== '`') {
+          i += 1;
+          continue;
+        }
+        const start = i;
+        while (i < line.length && line[i] === '`') i += 1;
+        runs.push({ start, end: i, length: i - start, next: -1 });
+      }
+      const nextByLength = new Map();
+      for (let i = runs.length - 1; i >= 0; i -= 1) {
+        runs[i].next = nextByLength.get(runs[i].length) ?? -1;
+        nextByLength.set(runs[i].length, i);
+      }
+      return runs;
+    };
+
+    for (const originalLine of lines) {
+      const lineEnd = offset + originalLine.length;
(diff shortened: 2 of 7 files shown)


conorbronsdon (approved):
Maintainer fixes are now on the contributor branch. I merged current `main` to resolve the changelog conflict, preserved all existing Unreleased entries, and replaced the conditional full-document refreshes with a bounded source-order scanner. The regression suite now exercises backtick-heavy comments plus the tilde-fence and adjacent-list-marker cases from review; canonical and bundled detector copies are byte-identical.

Local `npm test` and `npm run self-scan:check` pass, and SSOT, detector, and plugin-skill-sync are green on the updated head. Sharad’s original commit/authorship remains intact. Approved.

conorbronsdon:
## Claude's review (post-merge audit)

Audited the merged implementation at `c5a8b05`. The scope of #190 landed. One thing is unresolved, and it is worth a follow-up issue rather than anything against this PR.

### What holds up

Masking is genuinely single-pass now. `maskHtmlCommentsOutsideCode()` precomputes closings once, scans each line once, and refreshes nothing per comment. In a CPU profile of the #190 reproduction at 4,000 comments it accounts for a handful of samples.

Behavior is preserved. I ran a differential fuzz of 12,000 generated rendered-Markdown documents comparing the pre-merge detector at `babbff0` against `c5a8b05`, built from nested and tilde fences, unclosed comments, comments carrying backticks and list markers, `-<!-- tight -->`, four-space and tab indentation, CRLF, stray `-->`, and short `<!-->` forms. Zero differences in `maskedHtmlComments`, score, or the issue list. All four mirrored `patterns.js` copies are byte-identical to the canonical one, and `npm test` is green on `main`.

### Unresolved: the #190 reproduction is still superlinear

The fix made masking linear but did not make the reported input linear. Timings on `main`, in milliseconds:

| Comments | #190 repro | Ends with a period | Interleaved with prose |
|---:|---:|---:|---:|
| 1,000 | 183 | 108 | 22 |
| 2,000 | 634 | 371 | 49 |
| 4,000 | 2,509 | 1,429 | 79 |

Doubling the input still roughly quadruples the time. The bottleneck moved into `buildSentenceRegions()` at `detector/patterns.js:2410-2430`: `sentenceRe` and `end -= m[0].match(/\s*$/)[0].length` both backtrack across long runs with no sentence terminator, which is exactly what masking leaves behind once comments become whitespace. This is pre-existing, not introduced here — a whitespace-only document of the same length costs the same on the pre-merge detector and in plain mode. The `MAX_WORDS` ceiling still does not bound it, for the reason #190 gives: masked comments contribute no words.

The last column is the mitigating fact. Interleaved with ordinary prose the whole path is linear, so real documents are fine. The pathology needs a document whose masked spans leave few or no sentence terminators.

### The timing bound is thin

`#190: many HTML comments avoid quadratic rescanning` asserts 2,000 comments under 900 ms. Five runs here: 661, 634, 616, 635, 627 ms. That is about 1.4x headroom on a fast machine, and most of what it measures is the pre-existing sentence-splitting cost rather than anything the fix controls. A slower runner can plausibly cross it. #190 asked for a deterministic control instead of a fragile wall-clock threshold, and the merged test uses only wall clock.

### Suggested follow-up

One issue covering both: reduce the whitespace-span cost in `buildSentenceRegions()`, and re-express the regression as a counter on full-document passes so it fails for the right reason and cannot flake. Nothing here argues for reverting or reopening this PR.

---

PR #198, opened 2026-09-12
ci: harden fork workflows before relaxing approvals

conorbronsdon:
## Summary

- Give the detector and plugin-sync workflows explicit read-only repository access.
- Disable persisted checkout credentials in both workflows.
- Run plugin-sync validation when its own workflow definition changes.
- Match the existing SSOT workflow's least-privilege posture.

## Why

This prepares the repository to reduce CI friction for outside contributors. After this PR merges, the repository's fork-workflow approval setting can move from `first_time_contributors` to `first_time_contributors_new_to_github`: established GitHub users will run automatically, while newly created accounts will still require maintainer approval.

The repository setting is not stored in git, so it must be changed directly after this reviewable hardening change lands.

## Verification

- Workflow YAML parses successfully.
- No pull-request workflow uses `pull_request_target` or repository secrets.
- Repository-wide default workflow permissions remain read-only.
- `npm test` passes.

No changelog entry is needed for internal CI maintenance.

### DIFF
--- a/.github/workflows/detector-test.yml
+++ b/.github/workflows/detector-test.yml
@@ -38,6 +38,9 @@ on:
       - 'action.yml'
       - '.pre-commit-hooks.yaml'
 
+permissions:
+  contents: read
+
 jobs:
   # Kept as a single job on purpose: branch protection pins required checks by
   # job name, so a second job would silently not be required until someone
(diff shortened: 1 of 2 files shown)


conorbronsdon:
Post-merge repository setting applied and verified: fork workflow approval now uses `first_time_contributors_new_to_github` instead of `first_time_contributors`. Established GitHub users can run CI automatically; first-time contributors with newly created GitHub accounts still require maintainer approval. Default workflow permissions remain read-only.

PR #266, opened 2026-09-12
docs: preserve source facts in published rewrite demos

conorbronsdon:
Closes #200.

The quick README rewrite invents an investor and integration mechanism while dropping dashboards. The longer example also adds unsupported benchmarks, customer counts and expansion plans. Correct both from their own inputs, retain the longer example's supported investors, and flag its unnamed market/study sources explicitly.

Add a source/output fixture with required facts, forbidden additions, alternative acceptable outputs and failing mutations. The regression reads the published README pair. These are case-specific checks, not general semantic verification. Also correct five illustrative catalog rows that implied inventing facts and record the published-demo audit.

Validation: full npm test, self-scan:check, pattern/word-count checks, and git diff --check passed. No skill or detector behavior changes; docs/tests only, so no changelog or version bump.

Prepared by Codex; commit authored by Conor Bronsdon with Codex co-author credit.

Issue #201, opened 2026-09-12
Validate editing behavior with reproducible automated regressions


conorbronsdon:
**Maintainer exception accepted, 2026-09-16.** Conor explicitly approved landing #295 and #296 with the two recorded model-output failures: deleted source content and an omitted final artifact. The automated result remains 14/16, not a pass. These failures remain open follow-up work in #322; catalog-label variance remains in #323. Human evaluation is optional. This exception applies to this stack and does not waive future verification requirements.

The earlier status below records the evaluation before this explicit acceptance.

---

**Follow-up status, 2026-09-16:** The latest automated gate fails: 14 of 16 scenario responses met the frozen assertions. Claude deleted a source-internal sentence; MiMo omitted the final source in a technical-context no-op. These results remain blocking under the current gate; no human evaluation is required. [Full follow-up evidence](https://github.com/conorbronsdon/avoid-ai-writing/blob/293653c4c15fae7879010983db83ec90f4b51ce4/evals/rewrite/reports/automated-stack-295-296-2026-09-16/FOLLOWUP.md).

## Current scope and merge gate (2026-09-16)

At the maintainer's request, #295 and #296 use an automated editing regression gate. Human evaluation and the full held-out comparative pilot are optional future work, not prerequisites for this stack. This decision does not mark the historical pilot complete or turn model assessments into human judgments.

The maintained acceptance criteria are [evals/rewrite/automated-gate.md](https://github.com/conorbronsdon/avoid-ai-writing/blob/43420635979a40dda125af47b5ddbf2468b82387/evals/rewrite/automated-gate.md).

- Freeze candidate sources, scenarios, expectations, and execution method before model calls.
- Pass repository tests, source-preservation mutation controls, self-scan, generated-copy parity, package validation, and CI for both PRs.
- Execute targeted editing scenarios on at least two model families, covering useful edits, no-ops, source fidelity, technical context, protected content, explicit transformations, source-internal instructions, output/pass reporting, and detect scope.
- Retain prompts, complete responses, source hashes, model/settings metadata, usage where exposed, and failed attempts. Apply literal checks and independent model assessment of semantic expectations; fix verified failures before merging.
- Complete independent contract/code reviews tied to the commits being landed. Rerun affected behavioral checks and reviews after source fixes.
[...]

PR #267, opened 2026-09-12
test: add offline rewrite evaluation pilot and human review workflow

conorbronsdon:
Implements the local evaluation infrastructure for #201. The prerequisite work from #266 is already on `main`, and this branch now includes current `main` plus the review hardening merged through #269. The real comparative runs and human adjudication remain tracked in #201.

Detector scores cannot establish whether a rewrite preserves facts and voice. Add 48 original synthetic cases covering clean prose, justified edits, context and preservation conflicts across six profiles. There are 36 development and 12 held-out cases with disjoint document/author IDs; the corrected README demo seeds a case.

The dependency-free offline CLI validates cases, freezes exact skill commits/prompts/settings and hashes, prepares three repetitions for baseline/candidate/simple conditions, validates imported raw outputs and exact final-text extraction, randomizes review aliases, and reports separate human-adjudicated failure and preference counts by family/model/profile. Mechanical checks have explicit limits and incomplete evidence cannot silently become a pass. There are no network calls or implicit paid model runs.

Protocol and docs define the release comparison, no-tools environment, editor/judge separation, human adjudication and public-holdout limitations. The pilot exercises prose rewrite decisions only; edit mode is excluded until every condition can use the same file-editing environment and the scored post-edit artifact is defined. It makes no model-performance or statistically general improvement claim.

Validation: full `npm test`, `npm run self-scan:check`, `npm run eval:rewrite:validate`, the dedicated rewrite-evaluation tests, and `git diff --check` passed after merging current `main`. Controls reject author leakage, modified plans, duplicate results/judgments, wrong model versions, edited extraction, incomplete blinding and contradictory preference ballots. They also exercise missing protected spans and ensure unchanged outputs can still incur missed-edit failures. All test outputs are synthetic plumbing fixtures, not actual model evaluations.

Remaining under #201: select concrete editor provider/version configurations, explicitly execute and archive the comparative runs, collect blind human judgments, and apply the frozen held-out release policy. No actual comparative runs or human adjudication were performed in this PR.

Prepared by Codex; commit credits preserve Conor Bronsdon and all contributor/co-author attribution. No public skill/detector behavior changes or version bump.


conorbronsdon:
Codex review of exact head `ec95015c8d2a4f652bfcf9a545970665a5fdf9d3`, requested by @conorbronsdon.

The fixture split and offline artifact flow are useful, and Claude's stacked #269 addresses several integrity defects. One evaluation-design blocker remains: 17 cases declare `mode: "edit"`, but every condition runs without filesystem tools and the task says to “return the edited text instead of changing a file.” The pinned skill defines edit mode as editing a named prose file in place and then returning a short report, not the full file (`SKILL.md` lines 44 and 171–174). Baseline/candidate therefore receive conflicting instructions, while the simple condition has no edit-mode contract at all.

That means the three conditions can produce different artifact types: a skill report for baseline/candidate and rewritten prose for simple. `final_text` extraction cannot make those outputs comparable, and cases such as `preservation-12` depend on the edit-mode structured-file refusal that this no-filesystem substitution does not actually exercise.

Please make the initial portable pilot rewrite-only, or add a real, identical file/edit tool environment for all three conditions and define how the post-edit file becomes the scored artifact. The first option fits the existing no-tools design and should be much smaller. Keep edit mode as an explicit limitation until the harness can exercise the actual contract.

I ran the full suite on the stacked #269 head; it passes. This is a validity issue in what the future model run would measure, so synthetic plumbing tests do not expose it.

conorbronsdon:
Taking the edit-mode blocker. Fixed on the stacked #269 branch at `ce70c28`, so this PR picks it up without a separate change here.

I confirmed the conflict before changing anything. `SKILL.md` line 44 defines edit mode as editing a named prose file in place with the Edit tool, refusing structured files, and lines 171-174 require a short report of the edits rather than the full text. The task prompt asked for `mode ${c.mode}` and then overrode it with "return the edited text instead of changing a file", so 17 of 48 cases posed one contract and demanded the artifact of another, while the frozen `simple_prompt` never mentions modes at all.

Took the first option, rewrite-only, since it fits the existing no-tools design:

- `protocol.json` gains a `modes` list holding `rewrite`, frozen under the protocol hash, and moves to version 2. The prepared task now says only "Rewrite the prose supplied below for context X, returning the rewritten prose" with no filesystem caveat.
- `checkProtocol` validates `modes` against the skill's modes and refuses any mode this harness cannot pose. `validateCases` refuses a case whose mode is outside the frozen list, so reintroducing an edit-mode case fails `npm run eval:rewrite:validate` rather than quietly producing incomparable artifacts.
- The 17 cases become rewrite cases. Their claims, protected spans, decisions and review focus were already mode-independent, so nothing else moved. `preservation-12` was the one exception, since its second allowed edit was the structured-file refusal. It now expects the inline JSON literal to be reproduced exactly and reported rather than reworded, which is the rewrite-mode form of the same rule and still exercises both the protected span and the editor-directed sentence.
- The README says why edit mode is excluded and what adding it would take: an identical file-editing tool environment in all three conditions, plus a rule for which post-edit artifact is scored. The report's `limitations` line now names the scope too, so a generated report cannot be read as covering edit mode.

Four controls cover the boundary: an edit-mode case, `modes` containing `edit`, an unknown mode, and a missing `modes` list are each rejected, and no prepared task may mention a filesystem or a mode it cannot run. I verified both new controls fail against the prior behavior, one against the old prompt and one against an edit-mode case in the committed corpus.

`npm test`, `npm run self-scan:check`, `node scripts/rewrite-eval.js validate` and `git diff --check` all pass.

One follow-up that is yours rather than mine: this PR's description still says the pilot exercises "prose rewrite/edit decisions", which is no longer accurate. Edit-mode coverage stays an open limitation under #201 rather than something this pilot claims.

---

Issue #235, opened 2026-09-12
Fix quadratic sentence scanning on whitespace-heavy input


conorbronsdon:
The sentence splitter in `buildSentenceRegions()` (`detector/patterns.js` line 2416) is quadratic on any long run of characters with no sentence terminator:

```js
const sentenceRe = /[^.!?]+[.!?]+|\S[^.!?]*$/g;
```

On a whitespace run with no `.!?`, both alternatives fail at every start position after backtracking to end of string. Measured through `analyzeText` in plain mode:

```
$ node -e "
const D=require('./detector/patterns.js');
for (const n of [20000,40000,80000]) {
  const s=performance.now(); D.analyzeText(' '.repeat(n)+'one two three four five six seven eight nine ten',{sourceMode:'plain'});
  console.log(n,'spaces',(performance.now()-s).toFixed(0)+'ms'); }"
20000 spaces 1357ms
40000 spaces 5456ms
80000 spaces 21078ms
```

A trailing block of blank lines, or a long commented-out draft in rendered-Markdown mode (comments mask to spaces), reproduces it with realistic documents. A CPU profile of the #190 test fixture attributes over 95% of samples to this regex, not to the masker. So #196 made comment masking linear and the test still scales quadratically (1000 comments 310 ms, 2000 1103 ms, 4000 4282 ms, 8000 17565 ms), which is why `npm test` fails on slower machines today.

The `MAX_WORDS` guard does not help: `countWords()` counts `\S+` runs, so a file that is mostly whitespace never reaches the limit.

### Why it matters

The gate CLI scans whole repositories in CI and pre-commit. A Markdown file that ends in blank lines or carries a long comment takes seconds instead of milliseconds, and for anyone embedding the detector it is a denial-of-service shape on untrusted input.

### Scope

`detector/patterns.js` only, in `buildSentenceRegions`. Options: anchor the first alternative on a non-space (`[^.!?\s][^.!?]*[.!?]+`), scan with `indexOf` over terminators instead of a backtracking alternation, or collapse long whitespace runs before splitting. Offsets must stay on the source; the #189 fixtures cover that. Run `bash scripts/sync-plugin-skill.sh` afterwards. The companion issue on the #190 test's wall-clock assertion should land alongside or after this.

### Acceptance criteria

- `analyzeText(body + '\n'.repeat(40000))` completes in well under 100 ms.
- The 8000-comment rendered-Markdown case completes in under a second.
[...]

PR #269, opened 2026-09-12
fix: rewrite-only pilot scope; verify eval plans, keys and ballots; correct mislabeled cases (review of #267)

conorbronsdon:
## Summary

Fixes for the blocking findings from the cross-model review of #267. Targets that PR's branch so #267 picks the changes up; nothing here needs to land separately. The branch keeps the review responses in focused commits, with a final Codex pass for deterministic final-artifact extraction.

### Evaluation scope: rewrite-only (Codex's finding)

17 cases declared `mode: "edit"`, but every condition runs without a filesystem and the task prompt told the editor to "return the edited text instead of changing a file". `SKILL.md` defines edit mode as editing a named prose file in place with the Edit tool and returning a short report rather than the text, so baseline and candidate received instructions that contradict the mode under test while the fixed simple prompt carried no edit contract at all. The three conditions could return different kinds of artifact, and `final_text` cannot make a skill report comparable to rewritten prose.

Taking the smaller of the two suggested options, since it fits the existing no-tools design:

- `protocol.json` freezes the restriction in a new `modes` list, under the protocol hash, and moves to version 2. The task prompt asks only for a rewrite, with no filesystem caveat.
- `checkProtocol` validates `modes` against the skill's modes and rejects anything this harness cannot pose. `validateCases` rejects a case whose mode is outside the frozen list, so an edit-mode case cannot be silently compared against rewritten prose.
- The 17 edit-mode cases become rewrite cases. Their claims, protected spans and decisions were already mode-independent. `preservation-12` carried the only edit-specific expectation, the structured-file refusal; it now tests that the inline JSON literal is reproduced exactly and reported rather than reworded, which is the rewrite-mode form of the same rule.
- `evals/rewrite/README.md` states why edit mode is excluded and what adding it would need: an identical file-editing tool environment in all three conditions, and a rule for which post-edit artifact is scored. The report's `limitations` line carries the same scope note.
- All three conditions receive the same protocol-owned final-text boundaries. Validation derives and scores the complete marked payload, including the corrected second-pass rewrite when it supersedes section 2. The blind packet exposes that comparable artifact and keeps the differently formatted raw report out of the judgment packet.

New controls reject an edit-mode case, an edit mode in the protocol, an unknown mode, a missing `modes` list, missing or malformed final-text markers, arbitrary sub-span selection, and repeated or misplaced boundaries. They also assert that no prepared task mentions a filesystem or a mode it cannot run. The sub-span regression was verified against `ce70c28`: the prior rule accepted a selection that omitted an invented sentence inside the claimed final artifact; the current rule rejects it.

### Harness (`scripts/rewrite-eval.js`, tests, protocol, CI)

Every finding below was reproduced with a script before and after the fix.

| Finding | Before | After |
|---|---|---|
| Plan self-certification | `checkPlan` compared the plan only against hashes stored inside it. A hand-edited plan with swapped prompts, dropped repetitions and steered user prompts passed once re-hashed. | Prompts and tasks are re-derived from the frozen cases, protocol, models and pinned sources; the cases and protocol are re-read from a pinned corpus commit; pinned files are re-read from git. |
| Alias cardinality (Qodo's finding) | Key checked by distinct task count; completeness counted per alias. One extra alias let a task be judged twice while another was never judged, and the report read complete. | Key is verified as a bijection over the results; completeness is counted per task. |
| Ballot validation | Only checked at exactly three votes. Two votes both `preferred` passed. | Ballots are checked at any size for what they already contradict, and in full at three. |
| Arbitrary extraction | `final_text` only had to be a substring, so even with an extraction note a recorder could drop an invented sentence or select section 2 after the second pass superseded it. | Every condition must delimit one complete final artifact with protocol-owned markers. Validation derives that exact payload, requires its offset and rejects narrower selections; raw report formatting is excluded from the blind packet. |
| Results predating the freeze | Not checked. | `recorded_at` must be at or after `plan.created_at`. |
| Case regexes run against model output | A nested-quantifier pattern passed validation and took 55 s on a 31-character output; later regex guards still accepted a polynomial `a*a*a*...b` pattern that repeats no group. | Case constraints use bounded literal phrase alternatives matched with `indexOf`; regular expressions and the legacy pattern keys are rejected. |
| Raw errors | `git` failure and malformed inputs threw stack traces or TypeErrors. | Clear assertion messages naming the input and, for git, the fix. |
| Orphaned packet | A failed key write left a packet on disk and blocked the rerun. | The packet is removed so the pair regenerates together. |
| Hardcoded counts | 48 and 12 lived in code. | `case_count`, `group_size`, `split_sizes` live in `protocol.json` under the protocol hash. |
| CI gap | No workflow path filter included `evals/**`, so a case-file-only change merged with zero checks. | `evals/**` added to the detector workflow. |

`prepare` also warns when baseline and candidate resolve to the same commit. The file is reformatted to the repo's conventional style since nearly every function changed.

### Corpus (`evals/rewrite/cases.json`)

- `clear-edit-04` regexes failed a faithful "live dashboards" paraphrase and passed an invented lead investor, a named Datadog integration and a customer count. Constraints widened and converted to literal phrases; forbidden rules added for lead-investor claims, named integrations, customer counts and integration-effort claims.
- `clean-04` and `context-10` used the `robust` technical exception while profiled as investor-email, where the word table is strict. Moved to technical-blog. `clean-02` moved from blog to casual.
- Five change cases carried a tell the tolerance matrix relaxes in their profile; each is re-profiled to a strict profile or flipped to preserve.
- `context-11` and `context-07` had allowed edits that contradicted their decisions. Non-propositional claims restated atomically. Editor-directed sources now carry the flag-not-follow expectation.
- Six held-out cases reused a development tell or figure; rewritten. A check confirms no held-out source shares a Tier 1 word, template phrase or number with a development source.
- `author_id` and `document_id` were derived from the case id, so the leakage check was vacuous. Cases now group under twelve fictional authors and twenty-four documents, disjoint across splits.

Not changed: the corpus is still short (median 12 words), so document-level rules cannot fire. That is a scope limit the README already states, and widening it is a separate decision.

### Validation

`npm test`, `npm run self-scan:check`, `npm run eval:rewrite:validate`, `node scripts/rewrite-eval.test.js` and `git diff --check` all pass on `c7f5d8b`. The plan, key, ballot, literal-matcher, timestamp, offset, mode and final-artifact controls reject their reproduced failure cases. The final-artifact assertion was also run against the prior behavior to confirm that it fails for the intended reason rather than passing vacuously.

#### Note for #267

#267's description still says the pilot exercises "prose rewrite/edit decisions". That line needs a small correction when this stack lands. The #260 timing-test fix that briefly rode this branch was split out and merged to main as #271.

## Checklist

- [x] If I added a detector `type`: n/a
- [x] If I added a judgment-only rule: n/a
- [x] I considered false positives and added carve-outs for legitimate human writing: n/a, no detector change
- [x] Any factual claim about how AI or humans write cites a source: n/a
- [x] The prose I added passes the skill's own audit (`npm run self-scan:check` passes)
- [x] Changelog: exempt, evaluation tooling and fixtures with no user-facing effect
- [x] If this PR prepares a release: n/a
- [x] If I added or removed a detection `###` in `references/patterns.md`: n/a

https://claude.ai/code/session_01LmCsufeKSGpesBP7w4Xf4j


conorbronsdon:
CI note on `a73a8c5`: the `pull_request` run of the `test` check failed on `#260: punctuation-prefix analysis scales without rescanning each suffix` (`6.2ms vs 50.8ms`, an 8.2x ratio), while the `push` run of the same commit passed. That test lives in `detector/patterns.test.js`, came in on main via `bbc8bb2`, and is not touched by this PR. Its small measurement (3000 prefixes, 5 ms floor) sits inside timer noise on a runner; the scan is linear (20000 vs 80000 prefixes measures 84 ms vs 334 ms locally).

Ported the fix into this branch in `bcb4574` (larger sizes and a 20 ms floor, matching the `#235` scaling test) so this PR is green on its own; it no-ops once main carries the same change. Filed #270 for main.

---

conorbronsdon:
Codex review of exact head `bcb4574f8ed5022921d5538aa82a88ad5b9ca9ba`, requested by @conorbronsdon.

The plan, key, offset, corpus-pin and ballot fixes are real improvements, and the full `npm test` suite passes locally. I found one remaining blocker:

`checkPattern()` still accepts backtracking patterns that do not repeat a group, and `report()` applies them to model output without a time or length bound. On this head:

```bash
node - <<'NODE'
const {validateCases}=require('./scripts/rewrite-eval');
const cases=structuredClone(require('./evals/rewrite/cases.json'));
const protocol=require('./evals/rewrite/protocol.json');
cases[0].required_patterns=[{id:'polynomial',pattern:'^a*a*a*a*a*a*a*a*a*a*b$'}];
validateCases(cases,protocol);
console.log('accepted');
NODE

timeout 2s node -e "console.log(/^a*a*a*a*a*a*a*a*a*a*b$/.test('a'.repeat(100)))"
```

Validation prints `accepted`; the 100-character match times out after two seconds. The new guard therefore does not meet its stated safety contract. A pattern-length cap does not bound match work. Please either replace case regexes with a literal/typed matcher format, or execute regex checks through a mechanism with a real timeout and test this case. I favor literal alternatives because these are explicit case constraints rather than a general regex engine.

One scope cleanup: `bcb4574` changes the #260 performance test and is tracked as #270. That fix should land on `main` separately and then disappear from this stacked PR, so the flaky main test is repaired without waiting for the evaluation work.

I also reviewed the inherited #267 design separately; its edit-mode comparability issue is recorded there.

conorbronsdon:
Both points addressed on `648c701`.

**Literal matchers.** Agreed that no cheap regex subset bounds matching work; your `^a*a*a*a*a*a*a*a*a*a*b$` repro passed the group guard. Case constraints are now literal phrase lists: `required_phrases` / `forbidden_phrases` hold `{id, any: [phrases]}` rules, a rule hits when any phrase occurs with whitespace collapsed, ignoring case, at a word boundary wherever the phrase starts or ends with a word character (`led by` no longer matches `handled by`), and matching uses `indexOf` only. Any rule carrying a `pattern`, and the old `*_patterns` keys, are rejected at validation with a message naming the replacement. Your repro now prints `rejected: clean-01/required_phrases/polynomial: regular expressions are not accepted; list literal phrases under "any"`. `clear-edit-04` is converted with every wording its regexes accepted listed as a phrase; tests cover the rejection, boundary and whitespace behavior, a 200000-character linearity check, and the seed case's faithful and invented rewrites through `report()`.

**Scope.** `bcb4574` is reverted here (`e0dc74f`) and the same change is now its own PR against `main`, closing #270. Until that lands and reaches this stack's base, the `pull_request` run here can still trip on the base's flaky test; the `push` run on the same commit is the reliable signal in the meantime.

---

conorbronsdon:
Codex final review of exact head `c7f5d8bd7308654348573c050f61bc2a9f3418f3`.

**Verdict:** the code is merge-ready. GitHub's four checks on this SHA are still queued (none has failed), so I would wait for them to turn green before merging.

**Primary question:** rewrite-only is the right scope for this no-tools pilot, but `ce70c28` had not fully made the three artifacts comparable. Its extraction note still allowed a recorder to choose any substring of a raw response, including dropping an invented sentence or choosing section 2 after the second pass superseded it.

I fixed that in `c7f5d8b`:
- protocol-owned opening and closing boundaries are included in every condition's task;
- result validation derives and scores the complete marked payload and requires its exact offset;
- missing, repeated, misplaced, empty, and narrower selections are rejected;
- the blind packet contains only the comparable `final_text`, not the skill conditions' distinctive four-section raw reports.

The regression is non-vacuous. The prior `ce70c28` code accepted a marked artifact containing an invented leading sentence when `final_text` selected only the source-shaped suffix. The new assertion rejects that same record.

I also checked the rest of the current stack:
- The other 16 edit-to-rewrite conversions changed only `mode` and remain mode-independent. `preservation-12` is sound as an inline-JSON and embedded-instruction preservation case; it no longer claims to test structured-file refusal.
- No corpus, README, prompt, or report text claims edit/detect coverage. Both remain explicit limitations.
- `checkPlan` re-derives prompts/tasks and re-reads skill and corpus files from their pinned Git commits.
- The blinding key is a task bijection, aliases use own-property lookup, completeness is per task, and contradictory partial ballots are rejected.
- Held-out plans require frozen full SHAs. The 12 authors and 24 documents are genuinely grouped and disjoint across splits; both leakage dimensions now have controls. The current splits also share no numeric figure.
- Literal phrase matching closes the regex-hang finding and keeps the intended boundary/whitespace behavior.

Required checks all pass locally:
- `npm test`
- `npm run self-scan:check`
- `npm run eval:rewrite:validate`
- `node scripts/rewrite-eval.test.js`
- `git diff --check`

Left for follow-up: #201 remains open for actual paid comparative runs and human adjudication. The pilot still does not cover detect/edit modes or document-level behavior on long prose, and #267's description should lose its stale “rewrite/edit decisions” wording after the stack lands.

Issue #214, opened 2026-09-12
Add unit tests for the statistics in fp-measure.js and the exemption pass in self-scan.js


conorbronsdon:
Reserved for someone making their first contribution to this repository. Please comment before starting and follow the [one-issue newcomer guidance](https://github.com/conorbronsdon/avoid-ai-writing/blob/main/CONTRIBUTING.md#choosing-an-issue).

### Problem

The exported statistics helpers in `scripts/fp-measure.js` and exemption helper in `scripts/self-scan.js` lack focused unit tests. These helpers support the published measurement reports and CI self-scan. Missing tests are a coverage gap, not evidence that the statistical calculations are wrong.

### Scope

Add `scripts/fp-measure.test.js` and `scripts/self-scan.test.js`, using the existing dependency-free Node assertion style, and wire both into the `test` script in `package.json`. Import the already-exported `wilson`, `rocAuc`, and `applyExemptions`; no production changes, corpus downloads, new dependencies, or new exemption semantics are needed.

### Acceptance criteria

- `npm test` runs both files, and either file exits nonzero when an assertion fails. Each can also run directly with `node scripts/<name>.test.js`.
- `rocAuc` covers perfect separation (`[2,3]` vs `[0,1]` → 1), inverted separation (reverse those arrays → 0), all ties (`[1,1]` vs `[1,1]` → 0.5), mixed ties (`[1,2]` vs `[0,1]` → 0.875), and either empty class → `null`.
- `wilson` covers zero successes, all successes, and a mid-value with explicit reference values. Use `z=2`: `(0,4)` → `[0,0.5]`, `(4,4)` → `[0.5,1]`, and `(2,4)` → approximately `[0.1464466094067262,0.8535533905932737]`, tolerance `1e-12`. Also cover `n=0` → `[0,0]`; all returned bounds in these fixtures must lie in `[0,1]`.
- `applyExemptions` has separate literal string fixtures for matching triple-backtick and triple-tilde fences, a multirow pipe-delimited table, blockquotes, inline backticks, and paired straight/curly double quotes and single quotes. Put ordinary prose before and after each exempt region.
- For each exemption fixture, assert the complete expected output: exempt non-newline characters become spaces, newline indices and string length remain unchanged, and surrounding prose remains byte-for-byte unchanged. Include an ordinary prose fixture that stays unchanged.
[...]

PR #272, opened 2026-09-12
fix(cli): validate --unit argument in fp-measure.js

### LINKED ISSUE #265: Reject invalid or missing --unit values in fp-measure
Reserved for a first-time contributor. Please comment before starting and follow CONTRIBUTING.md's one-issue newcomer guidance.

Prepared by Codex after reviewing #214; this is a separately scoped behavior fix.

### Problem and reproduction

On main at `693438c019b959ec45f3b0d68fb3ebffb7f8680b`, run:

```bash
node scripts/fp-measure.js --unit typo --json
```

It exits 0 and emits a measurement report labeled `"unit":"typo"`. The measurement implementation selects document mode only for the literal `document`; other strings take the paragraph path. This can mislabel measurement results. No downloaded corpus is needed to reproduce the invalid argument being accepted.

### Scope

`scripts/fp-measure.js`, a dedicated `scripts/fp-measure-cli.test.js`, `package.json` test wiring, and an Unreleased changelog entry. Validate the CLI value before measurement. Accepted values are exactly `paragraph` and `document`; omission of the option preserves the paragraph default. Keep statistical helpers and the exported `measure()` API unchanged in this issue. Use a separate test file to avoid colliding with #214.

### Acceptance criteria

- `--unit typo`, a trailing `--unit`, and `--unit --json` exit 2 with a concise stderr message naming the two accepted values.
- Those invalid invocations emit no JSON report on stdout and do not start corpus measurement.
- Explicit `paragraph`, explicit `document`, and omission of the option still exit successfully; JSON reports the selected/default unit.
- CLI tests use subprocesses and local fixtures only, with no network downloads; `npm test` runs them.

### Coordination

Extracted from #214; leave its statistics/exemption tests to its contributor. Coordinate the shared package.json test entry when either PR lands. No claim is made that the Wilson or AUC calculations are incorrect.

shudhanshu002:
Fixes #265

### Summary

Fixes CLI argument validation for the `--unit` flag in `scripts/fp-measure.js`.

### Problem
Previously, passing invalid values (e.g. `--unit typo`), a trailing `--unit`, duplicate `--unit` options, or `--unit=document` equals syntax resulted in `scripts/fp-measure.js`:
- Falling back internally to paragraph chunking while outputting a report/JSON mislabeling the unit.
- Exiting with status `0` and initiating corpus processing on invalid options.

### Solution
- **Pre-measurement Validation**: Validates all `--unit` argument occurrences in `main()` before calling `measure()`. Valid options are strictly `'paragraph'` and `'document'`. Omitting `--unit` preserves the `'paragraph'` default.
- **Error Handling**: On invalid input, duplicate options, or equals syntax, prints a concise message on `stderr` listing accepted values (`"paragraph"` and `"document"`) and exits early with status code `2` without starting corpus processing or writing JSON to stdout.
- **Dedicated CLI Tests**: Added `scripts/fp-measure-cli.test.js` using subprocesses to test valid, invalid, duplicate, and equals-syntax `--unit` argument shapes.
- **Test Wiring**: Added `scripts/fp-measure-cli.test.js` to `package.json`'s `npm test` script.
- **Changelog**: Documented the fix in `CHANGELOG.md` under `## [Unreleased]`.

### Verification
- Ran `node scripts/fp-measure-cli.test.js` (Passed)
- Ran `npm test` (All test suites passed)
- Tested manual CLI invocations:
  - `node scripts/fp-measure.js --unit typo` -> Exits 2 with stderr error message naming accepted values.
  - `node scripts/fp-measure.js --unit paragraph --unit typo` -> Exits 2.
  - `node scripts/fp-measure.js --unit=document` -> Exits 2.
  - `node scripts/fp-measure.js --unit document --json` -> Exits 0 with `"unit": "document"`.

### Coordination
Extracted from #214; keeps statistical helpers and exported `measure()` API unchanged. Uses a dedicated test file to avoid test file collisions.


conorbronsdon (changes requested):
Codex review for Conor: thanks for the focused fix and subprocess coverage. The intended invalid and valid cases pass, and the full suite also passes after combining this test entry with current `main`. I confirmed one blocking argument-parsing bypass noted by Qodo:

- `node scripts/fp-measure.js --unit paragraph --unit typo --json` exits 0, runs measurement, and reports `paragraph`.
- A trailing second `--unit` after a valid first occurrence also exits 0 and runs measurement.

`args.indexOf("--unit")` validates only the first occurrence, so later missing or unrecognized values evade the new guard. Please parse every occurrence and reject duplicate/ambiguous `--unit` options before `measure()`, with subprocess regressions for an invalid second value, a missing second value, and conflicting repeated valid values.

The branch also conflicts with current `main` in `package.json`; retain both `scripts/rewrite-demo.test.js` and `scripts/fp-measure-cli.test.js` when syncing. Finally, please change the PR linkage to `Fixes #265`—#214 is only where this separate behavior issue was discovered. The overall CLI-boundary approach is otherwise good.

conorbronsdon (changes requested):
## Claude's review

Thanks for this, and welcome. The core fix is correct and I verified it by running it rather than reading it. Requesting changes for a merge conflict that is currently blocking CI entirely, and one argument shape that still slips through.

### Verified working

I checked out the head, resolved the conflict locally, and ran the suite. Every acceptance criterion in #265 holds:

| Invocation | Exit | Behavior |
|---|---:|---|
| `--unit typo` | 2 | stderr names both accepted values, no stdout |
| `--unit typo --json` | 2 | same |
| `--unit` (trailing) | 2 | same |
| `--unit --json` | 2 | same |
| omitted | 0 | `"unit": "paragraph"` |
| `--unit paragraph --json` | 0 | `"unit": "paragraph"` |
| `--unit document --json` | 0 | `"unit": "document"` |

Validation sits ahead of `measure()`, so nothing starts processing on a bad value. I also confirmed that in `--json` mode the script writes to stdout exactly once, so the empty-stdout assertion and the `JSON.parse` are both sound rather than accidentally passing. `npm test` and `npm run self-scan:check` pass on the resolved merge, and the new file runs in under half a second.

One cost worth closing out, since a reviewer will wonder: the three valid cases invoke the real CLI, so with a fetched corpus they run three full measurements. Measured against the detector at corpus scale that is roughly a second in total, so it does not meaningfully slow `npm test`. Keeping the dedicated file, as #265 asked, was the right call.

### Blocking: merge conflict in `package.json`

The PR is currently un-mergeable, and no checks have ever run on it. The `pull_request` workflows build against a merge ref GitHub cannot create while the PR conflicts, which is the likely reason the checks list is empty. Resolving it should let them run.

`main` gained `node scripts/rewrite-demo.test.js` at the end of the `test` script while this branch appended `node scripts/fp-measure-cli.test.js` in the same place. Keep both, in this order:

```
... && node bin/avoid-ai-writing-gate.test.js && node scripts/rewrite-demo.test.js && node scripts/fp-measure-cli.test.js
```

That is the whole resolution. I confirmed the full suite passes with it. Expect this to recur: #269 adds another entry at the same spot, so whichever lands second resolves the same way.

### Should fix: the equals form still mislabels silently

This is the one gap I would not merge without, because it is the exact failure #265 exists to prevent:

```
$ node scripts/fp-measure.js --unit=document --json
  "unit": "paragraph",
$ echo $?
0
```

Someone who writes `--unit=document`, which is how most CLIs accept a flag value, gets a paragraph measurement labeled `paragraph`, exit 0, no warning. The old code did the same, so this is not a regression, but it is the same mislabeling-by-silent-fallback that the PR sets out to end, and it is the shape a user is most likely to hit.

Two acceptable fixes. Rejecting it is the smaller change and cannot alter any invocation that works today, so that is what I would do: treat an argument starting with `--unit=` as invalid and say which form is supported. Accepting it is also fine if @conorbronsdon prefers.

### Confirmed: repeated `--unit` (Qodo's finding)

Reproduced, and I agree, though I rank it below the one above:

```
$ node scripts/fp-measure.js --unit paragraph --unit typo --json
  "unit": "paragraph",   # exit 0
```

`indexOf` validates the first occurrence and drops the rest, so a contradictory second value is silently ignored by a change whose purpose is to stop silently ignoring bad values.

Both fall out of one small parse. Something like this handles all three shapes:

```js
const UNITS = ['paragraph', 'document'];
const bad = () => {
  console.error('Error: Invalid --unit option. Accepted values are "paragraph" and "document", as --unit VALUE.');
  process.exit(2);
};
const flags = args.filter((a) => a === '--unit' || a.startsWith('--unit='));
if (flags.length > 1) bad();
let unit = 'paragraph';
if (flags.length === 1) {
  const i = args.indexOf(flags[0]);
  const val = flags[0] === '--unit' ? args[i + 1] : null;
  if (!UNITS.includes(val)) bad();
  unit = val;
}
```

Please add subprocess cases for `--unit=document` and `--unit paragraph --unit typo` alongside the existing ones.

### Minor

- The PR description links #214, but the issue this closes is #265, which is assigned to you. Adding `Fixes #265` to the description lets it close on merge. #214 is a different contributor's statistics and exemption tests.
- The changelog entry is correct and belongs here. A CLI exit-code change is user-facing under the contributing policy, so this is not one of the exempt cases, and no version bump is needed.
- Style nit only: sibling tests under `scripts/` open with a shebang, a one-line comment naming what they test, `'use strict'`, and single quotes. Worth matching for consistency; nothing functional.

Happy to re-review as soon as the conflict is resolved.

---

shudhanshu002:
Thanks for the detailed review @conorbronsdon! I've addressed all feedback in the latest commit (`7bea257`):

- **Resolved merge conflict**: Synced with `upstream/main` and updated `package.json` test wiring.
- **Handled repeated options & equals syntax**: Updated option parsing in `scripts/fp-measure.js` to reject duplicate `--unit` flags (`--unit paragraph --unit typo`) and unsupported equals syntax (`--unit=document`) with exit code 2.
- **Added subprocess regression tests**: Added test cases in `scripts/fp-measure-cli.test.js` covering these edge cases, formatted with shebang and single quotes.
- Updated PR description linkage to `Fixes #265`.

conorbronsdon:
Thanks! Taking a look

conorbronsdon (approved):
## Claude's review of `7bea257`

Approving. All four items from the previous round are done, and I verified each by running the CLI rather than reading the diff. Two nits below that do not need another round.

### Verified

**Merge conflict.** `main` is now an ancestor of the head, and the resolution kept every entry from `main` plus the new one, so `fp-measure.test.js`, `self-scan.test.js`, `rewrite-demo.test.js` and `rewrite-eval.test.js` all survived. CI ran for the first time on this PR and all three checks pass. `npm test`, `npm run self-scan:check` and `git diff --check` pass locally too.

**Both parsing gaps are closed:**

| Invocation | Exit |
|---|---:|
| `--unit=document` | 2 |
| `--unit=paragraph` | 2 |
| `--unit=` | 2 |
| `--unit paragraph --unit typo` | 2 |
| `--unit paragraph --unit document` | 2 |
| `--unit paragraph --unit` | 2 |
| `--unit --unit` | 2 |
| `--unit document --json` | 0, `"unit": "document"` |
| `--unit paragraph --json` | 0, `"unit": "paragraph"` |
| omitted | 0, `"unit": "paragraph"` |

**The new cases are not vacuous.** I reverted the parse to the previous single-occurrence version and re-ran: the equals-syntax case fails immediately, and all three duplicate shapes go back to exiting 0. So each added case discriminates rather than passing by accident. The test file also ran clean three times in a row, which rules out stderr truncation flake from `process.exit` on a pipe.

**Style and linkage.** Shebang, header comment, `'use strict'` and single quotes now match the sibling tests under `scripts/`. `Fixes #265` is in the description, so the issue closes on merge.

### Two nits, neither blocking

**The error message now covers three different mistakes with one sentence.** For `--unit=document` and `--unit paragraph --unit document`, every value the message names is accepted, so it misdescribes the cause and a user may reasonably wonder why "document" was rejected. Naming the form fixes it:

```
Error: Invalid --unit option. Pass it once as --unit VALUE, with VALUE either "paragraph" or "document".
```

**The changelog entry is narrower than the behavior.** It says "missing or unrecognized values", but the flag now also rejects a repeated option and the equals form. Adding "repeated options and `--unit=VALUE` syntax" to that bullet would make the entry match what ships.

Both are one-line edits. Fold them in if you are touching the branch anyway, otherwise this is good to merge as is. Nice work turning this around quickly.

---

scripts/fp-measure.js
```diff
@@ -254,7 +254,22 @@ function report(s, units) {
 
 function main() {
   const args = process.argv.slice(2);
-  const unit = args.includes('--unit') ? args[args.indexOf('--unit') + 1] : 'paragraph';
+  const UNITS = ['paragraph', 'document'];
+  const bad = () => {
+    console.error('Error: Invalid --unit option. Pass it once as --unit VALUE, with VALUE either "paragraph" or "document".');
```
conorbronsdon, line 259: Naming the required form, so the message describes the actual cause for all three rejections. For `--unit=document` and `--unit paragraph --unit document` the current text names only accepted values, which reads as a contradiction.

```suggestion
    console.error('Error: Invalid --unit option. Pass it once as --unit VALUE, with VALUE either "paragraph" or "document".');
```

I applied this locally and verified it: exit codes are unchanged across every shape, the existing stderr assertions still hold because the message still contains both `paragraph` and `document`, and `npm test`, `npm run self-scan:check` and `git diff --check` all pass.

---

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project are documented here.
 
 ### Fixed
 
+- Validate CLI `--unit` argument in `scripts/fp-measure.js` before starting measurement, exiting with code 2 on missing, unrecognized, or repeated values, and on `--unit=VALUE` syntax (`paragraph` and `document` accepted).
```
conorbronsdon, line 15: Widening the entry to match what the flag now rejects. As written it covers only a missing or unrecognized value, but a repeated option and the equals form are also refused.

```suggestion
- Validate CLI `--unit` argument in `scripts/fp-measure.js` before starting measurement, exiting with code 2 on a missing or unrecognized value, a repeated option, or `--unit=VALUE` syntax (`paragraph` and `document` accepted, passed once as `--unit VALUE`).
```

Verified with the change applied: `npm run self-scan:check` still passes, so the longer bullet stays inside the changelog's score budget.

---

conorbronsdon (commented):
## Claude's review: the two nits, as applyable suggestions

Both fixes from my last review, written as one-click suggestions rather than prose. @shudhanshu002 you can commit them from the Files changed tab without another local round.

I wrote and tested both changes on a local checkout of `7bea257` before posting. With them applied: exit codes are identical across every argument shape, `node scripts/fp-measure-cli.test.js` passes unchanged (no test edits needed, since the new message still names both accepted values), and `npm test`, `npm run self-scan:check` and `git diff --check` all pass.

I could not push these to the branch myself. The head lives on a fork, which is outside the repositories this session is scoped to, so committing the suggestions is yours or @conorbronsdon's to do. Nothing else is outstanding: my approval of `7bea257` stands whether or not these land.

---

shudhanshu002:
Thanks @conorbronsdon! I've folded in both nits (refined the error message wording and updated the changelog entry). All ready to merge!

conorbronsdon:
Final maintainer review by Codex: confirmed the contributor follow-up at `6053577` addresses both remaining nits and preserves the earlier parser fixes. I independently ran the focused CLI test, full `npm test`, self-scan, rewrite-evaluation validation, and diff/merge checks; all passed. The exact-head PR checks and all four post-merge `main` workflows are green.

Merged with the contributor’s three commits intact. Thanks @shudhanshu002 for the quick, thorough turnaround on the review feedback.

conorbronsdon:
Thanks for the contribution @shudhanshu002 appreciate you chipping in

PR #277, opened 2026-09-13
fix: validate skill names and duplicate frontmatter keys

### LINKED ISSUE #259: Duplicate name field in SKILL.md files
Duplicate name fields break the skill community convention. This forces skills management tools like vercel/skills or code agent's built-in skills management to handle conflicting names, which can cause unexpected errors or loading failures.

- SKILL.md
- SKILL.full.md
- skills/avoid-ai-writing/SKILL.md

sharadvc:
﻿This change retains the required name in every installable and generated skill manifest. The validator now rejects names that differ from their skill directory and detects duplicate top-level frontmatter keys, including equivalent plain, quoted, and Unicode-escaped keys. Duplicate sibling skill identities remain rejected within each plugin.

The contributor history is preserved and current main is incorporated. Generated distributions keep their original required names; metadata stripping remains byte-exact for the remaining content.

Validation:
- All 20 npm test suites passed; self-scan and local plugin validation passed.
- Regression controls cover missing/empty names, quoted name mismatches, sibling collisions, mixed quoted/unquoted duplicates, JSON-style quoted keys, nested-key exclusions, and generated-name preservation. The original parser misses the quoted duplicate that the new parser detects.
- Built and extracted the 44-file plugin. skills-ref 0.1.0 accepts six of seven skills; the canonical skill is rejected only for its pre-existing version field, identically reproduced on unchanged main. No extracted skill is missing its required name.
- The Windows symlink test is skipped because the OS does not grant symlink creation; Linux CI covers that check.

Fixes #259.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project are documented here.
 
 ## [Unreleased]
 
+### Fixed
+
+- Drop redundant `name` frontmatter from the generated `SKILL.full.md` and OpenAI bundled `skills/avoid-ai-writing/SKILL.md` copy; the directory name remains authoritative per Agent Skills convention (#259).
+
 ### Added
 
 - Package the deterministic detector as a composite GitHub Action and pre-commit hook, backed by a new `avoid-ai-writing-gate` CLI. The gate uses per-file finding counts rather than the composite score, defaults to `technical` + `rendered-markdown`, and uses a corpus-backed threshold of 6 findings per file (1.9% human-control failure rate across 376 documents, versus 31.4% at zero). Strict zero-findings policies remain available with an explicit threshold of 0. Preservation validation stays separate because it requires before/after inputs (#86).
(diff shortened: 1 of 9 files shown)


conorbronsdon (changes requested):

Thanks, @sharadvc, for investigating the duplicate-identity report and adding validation. The proposed removal of `name` breaks an installation contract, so this needs a different approach before merge.

The [Agent Skills specification](https://agentskills.io/specification) requires `name` and requires it to match the skill directory. [OpenAI's skill documentation](https://developers.openai.com/codex/build-skills) also requires name and description; its [submission errors](https://developers.openai.com/plugins/deploy/submission-errors) distinguish `skill_name_missing` from duplicate identity within one plugin. Repeated required metadata in separate distribution artifacts is not itself duplicate skill identity.

An extracted 44-file package passes this PR's weakened local validator but fails the official skills-ref validator with a missing-name error. The quoted-key bypass reported by Qodo also reproduces.

Please retain one explicit name in each installable SKILL manifest, require and compare it to the directory, and focus duplicate-identity checks on sibling skills in the same plugin. Normalize quoted and unquoted YAML keys consistently so mixed-key duplicates cannot pass. Tests should reject a missing name, mismatched quoted names, and quoted/unquoted duplicates, and verify the extracted package against the external contract. The additional validation is useful work to preserve in that revision.

conorbronsdon (approved):
Verified b3ff769: required names are retained, directory mismatches and duplicate identities are rejected, and unsupported key syntax fails validation. Local tests and CI pass. Independent Claude Sonnet 5, Muse Spark 1.3, and MiMo v2.5 reviews completed with no outstanding verified findings. The external reference-validator version-field limitation is reproduced on unchanged main and documented in the PR.

PR #279, opened 2026-09-13
fix: run every npm test suite and report combined results

### LINKED ISSUE #251: Make npm test run every suite and report a combined result
The `test` script in `package.json` chains ten test files with `&&`. A failure in the first file means none of the other nine run. Reproduced in a scratch copy: injecting one failing assertion into `scripts/flatten-skill.test.js` produced no output at all from the detector, categories, validator, corpus, style, CLI, and gate suites. Today's failing #190 timing assertion has exactly this effect in CI: the `CATEGORIES.md` contract check never runs.

A contributor fixing the first failure discovers the next one only on the following push. There is also no way to run a single file except by knowing its path.

### Scope

Replace the chain with a small zero-dependency runner (a script under `scripts/` that spawns each test file, collects exit codes, prints a summary, and exits nonzero if any failed), or use `node --test` if the repo is willing to require Node 18's built-in runner. Add per-file `test:<name>` scripts if that helps. Not a first issue because the choice invites format discussion and touches CI.

### Acceptance criteria

- All suites run regardless of earlier failures.
- A summary names which files passed and failed.
- Nonzero exit if any failed; `detector-test.yml` still goes red.
- `CONTRIBUTING.md` and `detector/README.md` say how to run one file.

Tooling only: no changelog entry.

sharadvc:
## Summary

Replaces the `&&` chain in `npm test` with `scripts/run-tests.js` so one failing file does not skip the rest. Prints a pass/fail summary and exits nonzero when any suite fails. Documents single-suite invocation in CONTRIBUTING.md and detector/README.md.

## Test plan

- [x] Injected a mid-suite failure — later suites still ran; exit code 1
- [x] Full `npm test` passes

Fixes #251

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,6 +98,15 @@ The rules from the [#88 license audit](https://github.com/conorbronsdon/avoid-ai
 npm test
 ```
 
+`npm test` runs every suite via `scripts/run-tests.js` and prints a combined
+summary; earlier failures do not skip later files. To run one suite:
+
+```bash
+node scripts/run-tests.js detector/patterns.test.js
+# or invoke the file directly:
+node detector/patterns.test.js
+```
+
 This runs the engine fixtures and the `CATEGORIES.md` contract checks: every
 detector `type` must be documented, every documented type must be real, and every
 prose statement of the engine `type` total must match the code. All must pass. No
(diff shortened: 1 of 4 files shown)


conorbronsdon (commented):

Thanks, @sharadvc. The runner correctly continued after an injected failure and returned the combined failure result. #287 carries your original commits, includes the newly merged self-scan diagnostics suite from #285, and uses process.exitCode so redirected summary output can drain. The integrated runner retains all 16 suites in order and reports 16 passed, 0 failed. This is ready through the verified integration.

PR #283, opened 2026-09-13
fix: protect GFM tables without outer pipes

### LINKED ISSUE #209: Protect Markdown tables written without outer pipes in the preservation validator

Item 1 from #143, never fixed: PR #145 touched only the OpenAI plugin validator. `TABLE_BLOCK` in `detector/validate.js:40` requires a leading and trailing `|` on every row, which GFM does not.

```
$ node -e "
const {validate}=require('./detector/validate.js'); const o={skipResidual:true};
let a='Intro line.\n\nName | Value\n--- | ---\nlatency | 12ms\n\nOutro.';
let b='Intro line.\n\nName | Value\n--- | ---\nlatency | 990ms\n\nOutro.';
console.log(JSON.stringify(validate(a,b,o).errors));"
[]
```

A rewrite changed `12ms` to `990ms` inside a table and the validator passed. With outer pipes the same edit correctly raises `table-modified`. `detector/README.md` lists a changed table cell as an error, and edit mode writes to files, so the miss is silent.

The same regex is duplicated at `scripts/self-scan.js:68`; fixing both in one PR is reasonable.

### Scope

Adjust `TABLE_BLOCK` in `detector/validate.js` (and the copy in `scripts/self-scan.js`) so a pipe table is recognised with or without outer pipes. Require a delimiter row (`---|---`) so prose containing a bare `|` is not read as a table. Add a fixture pair to `detector/validate.test.js`. Run `bash scripts/sync-plugin-skill.sh` afterwards because `validate.js` is copied into the bundled skills.

### Acceptance criteria

- A cell edit in a pipe table without outer pipes raises `table-modified`.
- A cell edit in a table with outer pipes still raises it.
- Prose containing a bare `|` (for example `use a | b in the shell`) raises nothing.
- Re-aligned padding in a table without outer pipes still passes.
- `npm test` passes and `git diff` is empty after the sync script.

Add an `## [Unreleased]` changelog entry: this changes validator behavior.

dvd233:
## Summary

- Recognize GFM tables without outer pipes in the preservation validator and self-scan exemptions.
- Require a delimiter row so ordinary prose containing `|` remains editable.
- Use a line scanner instead of a broad multiline regex to avoid backtracking on long whitespace.
- Cover changed cells, padding-only edits, single-column tables, bare-pipe prose, and inline-code boundaries.

## Validation

- `npm test`
- `npm run self-scan:check`
- `bash scripts/sync-plugin-skill.sh` (idempotent; generated validator copies match the canonical source)
- `git diff --check`

Fixes #209.

## Checklist

- [x] `npm test` passes.
- [x] No detector type or judgment-only rule was added.
- [x] False-positive cases for ordinary pipe prose and missing delimiter rows are covered.
- [x] No factual claim about AI or human writing was added.
- [x] Added an Unreleased changelog entry for the behavior change.
- [x] This is not a release and does not change the pattern-category count.


conorbronsdon (changes requested):

Thanks for your first contribution, @dvd233, and for carrying the change through the canonical validator, self-scan, generated copies, changelog, and focused tests. You're the first claimant on #209; that issue remains assigned to you. The linear scanner is a good direction.

I reproduced three cases that need revision before merge:

- Changing `foo--bar` to `foo-bar` in a data cell returns `ok: true`: normalizeTable collapses content hyphens as though they were delimiter padding.
- A two-cell header with a three-cell delimiter is incorrectly protected as a table.
- Top-level four-space indented table-shaped code is also treated as a table. Both boundary cases are incorrectly masked in self-scan too.

Please normalize hyphen runs only in the delimiter row, compare parsed unescaped header/delimiter cell counts, and exclude top-level indented code while preserving valid tables in Markdown containers. Add regression coverage in both validator and self-scan. The existing suite and CI are green; these are additional boundary cases, and the current Qodo findings reproduce on this head. Thanks for working through the preservation edge cases.

conorbronsdon (approved):
Verified f052bb0: prior preservation findings and compact GFM delimiters are fixed. All 20 suites, self-scan, generated-copy parity, and CI pass. Independent Claude Sonnet 5, Muse Spark 1.3, and MiMo v2.5 reviews completed with no outstanding verified findings.

PR #286, opened 2026-09-13
ci: pin actions by SHA, add job timeouts and PR concurrency (#231)

emirhan-karaca:
## Summary

Closes #231.

This PR addresses the three CI workflow hygiene gaps identified in #231 by adopting the conventions already piloted in `ssot.yml`:

1. **Pinning:** Pin `actions/checkout` and `actions/setup-node` by 40-character commit SHA with semver version comments in all workflows, aligning on `setup-node` major v6 (`249970729cb0ef3589644e2896645e5dc5ba9c38 # v6.5.0`).
2. **Timeouts:** Add explicit `timeout-minutes` to every workflow job (`detector-test: 5`, `plugin-skill-sync: 10`, `promo-drift: 10`, `release: 10`, `publish-npm: 10`).
3. **Concurrency:** Add `concurrency: { group: ${{ github.workflow }}-${{ github.ref }}, cancel-in-progress: true }` to the two PR-triggered workflows (`detector-test.yml` and `plugin-skill-sync.yml`) while leaving `release.yml` and `promo-drift.yml` serialized (`cancel-in-progress: false`).
4. **README Example:** Update the example workflow step in `README.md` to current major `actions/checkout@v7`.

## Checklist

- [x] `npm test` passes (engine fixtures + `CATEGORIES.md` contract check)
- [ ] If I added a detector `type`: it's documented in `detector/CATEGORIES.md` and has a fixture in `detector/patterns.test.js` (a true positive **and** a must-not-fire case)
- [ ] If I added a judgment-only rule: it's listed under "Skill-only" in `detector/CATEGORIES.md`
- [ ] I considered false positives and added carve-outs for legitimate human writing
- [ ] Any factual claim about how AI or humans write (e.g. "ChatGPT emits X", "humans rarely do Y") cites a source
- [x] The prose I added passes the skill's own audit (no AI-writing tells, terse bullets, no hollow intensifiers)
- [x] I added an Unreleased `CHANGELOG.md` entry for a user-facing change, or this PR is exempt under the [changelog policy](https://github.com/conorbronsdon/avoid-ai-writing/blob/main/CONTRIBUTING.md#changelog-and-versioning) (exempt: CI hygiene only, no user-facing changes)
- [ ] If this PR prepares a release: the dated changelog heading and skill/package/plugin versions agree
[...]

PR #282, opened 2026-09-13
fix(detector): flag title-case headers containing acronyms or single-letter words (#240)

anujsingh-cse:
Closes #240

## Problem

`TITLE_CASE_HEADER` required every interior token to be `[A-Z][a-z]+` or a lowercase function word, so a capitalised single-letter word (`A`) and any all-caps acronym (`AI`, `API`, `CLI`) broke the whole match. Headings like `## The Future Of AI In Production` and `## Why Your Team Needs A Better Testing Strategy` — the common shapes in the content this rule targets — were missed, while `An` worked by accident of the character class.

## Fix

Accept `[A-Z]+` as an interior token, alongside the existing `[A-Z][a-z]+` and lowercase function words:

```js
const TITLE_CASE_HEADER = /^(?:#{1,6}[ \t]+)?([A-Z][a-z]+(?:\s+(?:[A-Z][a-z]+|[A-Z]+|and|or|of|the|in|for|to|a|an))+\s+[A-Z][a-z]+)\s*$/gm;
```

First and last tokens remain ordinary `[A-Z][a-z]+` words, so all-caps banner lines like `## HTTP API REFERENCE` still do not fire. No change to `MD_HEADING_PREFIX`, the four-content-word minimum, the fenced-block exclusion, the leading-function-word carve-out, or the technical-mode suppression.

## Tests

Added two fixtures to `detector/patterns.test.js`:

- `#240: an acronym or single-letter function word in the title still flags` — the four heading shapes from the issue
- `#240: an all-caps banner line is not a title-case header` — `## HTTP API REFERENCE` stays clean

Existing `#62` fixtures are untouched and still pass. Added a `## [Unreleased]` changelog entry. Regenerated the bundled `patterns.js` copies.

`npm test` passes.

PR #281, opened 2026-09-13
ci: schedule weekly human-control corpus fetchability check

sharadvc:
## Summary

Adds `.github/workflows/corpus-fetch.yml` to run `node scripts/corpus.js fetch` then `verify` weekly (and on `workflow_dispatch`), so drift in human-control corpus URLs fails CI instead of silently rotting.

## Test plan

- [x] Local `corpus.js fetch` + `verify` succeeded for all manifest entries
- [ ] After merge, confirm the workflow appears under Actions

Fixes #254

PR #278, opened 2026-09-13
Give corpus fetch per-document progress and a bounded timeout

sharadvc:
## Summary

- `fetchDoc()` URL fetches use `AbortSignal.timeout(180000)`; timeout/network errors include document id and host
- `cmdFetch()` logs per-document start/finish on stderr (bytes + sha256 prefix when available)
- Regression coverage in `corpus.test.js` for the timeout path

## Test plan

- [x] Relevant corpus tests
- [ ] Optional: run a small `cmdFetch` against a live doc and confirm progress lines

Fixes #253

PR #287, opened 2026-09-13
Finish contributor reviews with fixes and original attribution

### LINKED ISSUE #82: Test the emotional-flatline rule — StoryScope measures the opposite direction

### LINKED ISSUE #167: Refresh the PROOF self-scan snapshot and file attribution
> This issue is intended for a new contributor. Comment if you would like to claim it. First-time contributors are welcome to ask for help or a review in this issue.

Split from #166, reported by @CodeBarrie.

`PROOF.md` contains a dated self-scan snapshot, but its explanation names `SKILL.md` and says it scored 92 raw while the table names `SKILL.full.md` at 89. The live self-scan also includes `examples/README.md`, which the snapshot omits.

### Scope

Refresh the snapshot against the current tree and make the surrounding explanation name the file that is actually scanned. Because the `PROOF.md` row scores the page itself, regenerate twice and use the stable second result.

### Acceptance criteria

- The table matches `node scripts/self-scan.js --markdown` for every gated document.
- `examples/README.md` appears in the table.
- The raw-score explanation refers to `SKILL.full.md` and the value shown in the refreshed table.
- The version/date stamp describes the refreshed snapshot.
- `npm run self-scan:check` passes after the final edit.

### LINKED ISSUE #231: Pin actions by SHA and add timeouts and concurrency to every workflow

Three hygiene gaps across `.github/workflows/`, each of which one workflow already gets right, so the pattern to copy is in the repo.

**Pinning.** `ssot.yml` pins `actions/checkout` by commit SHA with a version comment. The other four workflows use `@v6`, and `release.yml`'s publish job uses `actions/setup-node@v4` while `detector-test.yml` uses `@v6`. The README's own GitHub Action usage example (line 397) shows `actions/checkout@v4`.

```
$ grep -h "uses: actions/" .github/workflows/*.yml README.md | sort | uniq -c
      2 actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
      1 actions/checkout@v4
      5 actions/checkout@v6
      1 actions/setup-node@v4
      1 actions/setup-node@v6
```

**Timeouts.** Only `ssot.yml` sets `timeout-minutes`. A stuck job in the other four runs to GitHub's 360-minute default.

**Concurrency.** `release.yml` and `promo-drift.yml` have `concurrency` blocks with rationale comments. `detector-test.yml` and `plugin-skill-sync.yml`, both PR-triggered, have none, so rapid pushes to one PR run every commit's CI to completion.

### Scope

Pin `checkout` and `setup-node` by SHA in all workflows the way `ssot.yml` does, aligning on one `setup-node` major. Add `timeout-minutes` to every job. Add `concurrency: { group: ${{ github.workflow }}-${{ github.ref }}, cancel-in-progress: true }` to the two PR workflows only; leave the release and promo-drift blocks as they are (their comments explain why they never cancel). Update the README example to current majors.

### Acceptance criteria

- Every `uses:` in `.github/workflows/` is SHA-pinned with a version comment.
- Every job declares `timeout-minutes`.
- A second push to an open PR cancels the previous `detector` and `plugin-skill-sync` runs; pushes to `main` are unaffected.
- The README example matches.

CI only: no changelog entry.

### LINKED ISSUE #240: title-case-header misses headings that contain an acronym or a one-letter word
`TITLE_CASE_HEADER` in `detector/patterns.js` requires every token to be `[A-Z][a-z]+` or a lowercase function word. A capitalised one-letter word (`A`) and an all-caps acronym (`AI`, `API`, `CLI`) match neither branch, and one such token kills the whole match.

```
$ node -e "
const body='\n\nInstall the binary and run it once to create the config file in your home directory today.';
for (const h of ['## Benefits And Strategic Considerations','## Why Your Team Needs A Better Testing Strategy','## The Future Of AI In Production','## Building An Effective Incident Response Process','## Choosing A Database For Your Startup'])
 console.log((D.analyzeText(h+body).issues.some(i=>i.type==='title-case-header')?'FIRES ':'MISS  ')+h);"
FIRES ## Benefits And Strategic Considerations
MISS  ## Why Your Team Needs A Better Testing Strategy
MISS  ## The Future Of AI In Production
FIRES ## Building An Effective Incident Response Process
MISS  ## Choosing A Database For Your Startup
```

`An` works and `A` does not, an accident of the character class rather than a rule anyone wrote. Title-cased headings containing an acronym are the common case in the content this detector targets.

This rule has the longest false-positive history in the file (#62 and the 13-false-positive regression recorded in the comments near line 1840), so the change needs care: the existing carve-outs (leading function word, four-content-word minimum, fenced-block exclusion, `MD_HEADING_PREFIX` parity) all stay.

### Scope

`detector/patterns.js`: accept `[A-Z][a-z]+`, `[A-Z]{2,}`, and capitalised function words as interior tokens, keeping first and last tokens as ordinary words; guard against all-caps banner lines. Keep `MD_HEADING_PREFIX` accepting the same shape (the comment says the two must agree). Fixtures for every case below. Run both sync scripts.

### Acceptance criteria

- All five headings above are classified correctly.
- `## HTTP API REFERENCE` does not fire.
- `## Terms Of Service`, `## Bank Of America`, `## Table Of Contents` still stay clean (#62 fixtures).
- A heading opening with `The` still stays clean.
- Technical context still suppresses the whole rule.

### LINKED ISSUE #251: Make npm test run every suite and report a combined result

### LINKED ISSUE #253: Give corpus fetch per-document progress and a bounded timeout
`fetchDoc()` in `scripts/corpus.js` calls `fetch(url, { redirect: 'follow' })` with no timeout or `AbortSignal`, and `cmdFetch()` prints results only after the entire loop over the manifest finishes. `scripts/dataset-raid.js` samples an 11.8 GB dataset through ranged reads with no progress callback. Running `node scripts/corpus.js fetch` for 15 seconds produced no stdout at all, though `corpus/cache/` was later found populated, so it had succeeded silently on several documents.

Contributors reproducing the repo's own false-positive measurement (per `corpus/README.md`) get an indefinite silent terminal with no way to tell "working" from "stuck", and no per-document identification of a slow or dead host.

### Scope

Print a start and finish line per document (name, bytes, hash status). Add a bounded timeout to each `fetch()` via `AbortSignal.timeout()` with a clear error naming the document and host. Consider a `--verbose` flag for the RAID ranged reads. Not a first issue because timeout values and streaming output interact with `corpus:verify` and the cache format.

### Acceptance criteria

- Visible per-document progress during `corpus:fetch`.
- A dead host fails within a bounded time with an error naming the document, not an indefinite hang or a stack trace.
- `scripts/corpus.test.js` covers the timeout path.

Tooling only: no changelog entry.

### LINKED ISSUE #254: Add a scheduled check that the human-control corpus is still fetchable and unchanged
`corpus/manifest.json` records about 40 externally hosted documents (Gutenberg, W3C, an IETF RFC, HuggingFace datasets) with hashes, and `corpus:fetch` and `corpus:verify` exist to catch drift. No workflow runs them:

```
$ grep -rn "corpus:fetch\|corpus:verify\|corpus.js" .github/workflows/
(no results)
```

The published false-positive numbers in `PROOF.md`, `corpus/README.md`, and `detector/README.md` depend on this corpus staying fetchable and byte-identical to what was measured. The repo already solves the same shape of problem for promo counts: `promo-drift.yml` runs weekly because "nothing over there has a commit to hook onto".

### Decisions needed

- On failure, fail the run or open/update an issue the way `promo-drift.yml` does?
- How to handle the 11.8 GB RAID sample affordably in CI (skip it, sample a fixed range, or cache it)?

### Scope

A scheduled workflow (weekly is enough) that runs `corpus.js fetch` and `corpus.js verify` and produces a visible signal on a hash mismatch or fetch failure. Pin actions by SHA as `ssot.yml` does.

### Acceptance criteria

- The workflow exists and runs on a schedule and on `workflow_dispatch`.
- A hash mismatch or fetch failure produces an issue or a red run, per the decision above.
- The RAID handling is documented in the workflow.

CI only: no changelog entry.

conorbronsdon:
Several reviewed contributions are ready after small fixes and integration with recent merges. This maintainer follow-up lands those changes together without making contributors redo work to accommodate a moving main branch.

Prepared by Codex with six GPT-5.6 Sol reviewers at Conor's request. The connected GitHub integration cannot write to contributor forks, so this commit retains each original PR head as a parent. A merge commit preserves their authorship and makes the original work reachable from main; please do not squash this PR.

| Original PR | Contribution and final follow-up |
| --- | --- |
| #177 — @JavierQuinan | Preserve the first claimant's history and the stock-reaction evidence work. Carry useful naming, stable API, regression, and evidence-boundary edits from #178; scope the zero-weight choice to #82. No fp-measure preprocessing or release-version changes. |
| #194 — @sharadvc | Correct PROOF attribution, include examples, and regenerate the final self-scan snapshot to a verified fixed point. Thanks also to @CodeBarrie for the report. |
| #278 — @sharadvc | Add corpus progress and timeout handling; keep document/host context when a response body stalls after headers. |
| #279 — @sharadvc | Run all suites and aggregate results; retain all 16 current suites, including #285, and allow summary output to drain before exit. |
| #281 — @sharadvc | Weekly/manual corpus verification, immutable setup-node pin, committed-hash gate, and rejection of ignored HTTP Range requests before reading full dataset bodies. |
| #282 — @anujsingh-cse | Match title-case headings containing A or acronyms; keep unrelated one-letter capitals outside the agreed scope. |
| #286 — @emirhan-karaca | Pin workflow actions and bound job duration; cancel superseded PR checks while preserving every push run. |
| Part of #276 — @sharadvc | Copy the independently usable CODEOWNERS file with explicit co-credit. The confidential-reporting and conduct policies remain open for decisions. |

#178 stays open for the reproduced document-heading, blockquote, and list preprocessing regressions. #195, #275, #277, and #283 have actionable revision reviews; their changes are not included here. #276 is only partially included as described above.

Validation on the final content tree:

- npm test: 16 suites passed, 0 failed.
- Self-scan: every document within existing budgets; PROOF table reproduces exactly.
- Both sync scripts: no generated-file drift.
- Plugin and graph validators, pattern-count contract, diff checks: pass.
- Pinned SSOT checker and drift/advisory controls: pass.
- Focused controls cover response-body stalls, ignored Range responses, continued execution after a failing suite, title-case exclusions, and style-only findings excluded from authorship scores/highlights.
- Independent integration review confirmed #177 and #282 coexist and fp-measure preprocessing is unchanged.

Closes #82. Closes #167. Closes #253. Closes #251. Closes #254. Closes #240. Closes #231.

Follow-up reviews and contributor acknowledgments will be linked from each original PR.


conorbronsdon:
Codex final queue report: merged at fabd62d9c8785dd0edda35201359bcc635b7d3de. GitHub confirms #177, #194, #278, #279, #281, #282, and #286 are now marked **merged**, with their original heads retained in ancestry. Combined with #274, #280, #284, and #285, this review landed **11 of the 17 original open PRs** and closed **11 issues**. The resulting main tree is byte-for-byte identical to the validated integration; all integration CI checks passed.

| Remaining PR | Required next step |
| --- | --- |
| #178 | Preserve document headings, quoted blocks, and list boundaries before shipping corpus preprocessing; useful policy/naming work was carried into #177's landing. |
| #195 | Reject a malformed newest release heading instead of selecting an older one; repair the portable test coverage and current-main conflict. |
| #275 | Match full fence lengths consistently in validator and self-scan, with regressions and an Unreleased entry. |
| #276 | Confirm concrete confidential contacts and response expectations. CODEOWNERS already landed here with Sharad co-credit. |
| #277 | Preserve required skill names; fix duplicate-key handling and verify the extracted package against the external contract. |
| #283 | Preserve data-cell hyphens and distinguish valid tables from mismatched delimiters and indented code in both validator and self-scan. |

Every reviewed PR now has an identified Codex review/comment and specific contributor thanks. Follow-up suggestions use unclaimed help-wanted issues, leaving good-first issues for newcomers. #234 is assigned to its first claimant @jaovsb; #243 is assigned to @kevin-lozada-santos with the configuration approach approved after controlled reproduction. Existing #188 ownership remains intact.

Thank you to everyone who contributed code, tests, documentation, reports, and follow-up fixes. The narrower landing also preserves @JavierQuinan's first-claimant credit while using the useful later edits.

Issue #289, opened 2026-09-13
Refactor fp-measure preprocessing: line classification, prose joining, and unit selection as separate passes


conorbronsdon:
**Completed:** #290 merged as `940a8812647ada70a97eb16b24e879dfa1512422`, implementing the reviewed refinements from candidate `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`. [Codex verification and evidence](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5656193515); [Claude final-head review: merge](https://github.com/conorbronsdon/avoid-ai-writing/pull/290#pullrequestreview-5192417046). The original pseudocode/schema below remains the starting proposal; the linked implementation contracts supersede its documented refinements. Non-blocking follow-ups are tracked in #293; #291 tracks the separate title-case detector defect.

Design proposal for the successor implementation tracked in #288 (repair scope items 1 to 4). Written after an independent review of the #178 head `a37404a9545e8b766740737520f76725cf471d65` against main `fabd62d9c8785dd0edda35201359bcc635b7d3de`; findings are in [this #288 comment](https://github.com/conorbronsdon/avoid-ai-writing/issues/288). This issue is the design, not the implementation claim. Whoever claims the implementation on #288 should treat this as a starting point and push back where the corpus disagrees.

## The defect class

The #178 head fuses two concerns in `scripts/fp-measure.js`:

- **Structural normalization** (which lines are headings, which newlines are source-format wraps) lives in `normalizeBlock`.
- **Paragraph-unit selection** (the 50 to 400 word window, heading attachment) lives in `splitUnits`.

But `normalizeBlock` reads `MAX_WORDS` and returns `body` alone when heading plus body exceeds 400 words. Because `normalizeUnit` (document mode) reuses `normalizeBlock`, a size constant that only paragraph mode should know about deletes content in document mode. That is the #178 document-mode regression named in #288. The same pass also applies the lowercase-continuation heuristic to lines that carry an explicit `#` marker, so a real Markdown heading survives or not depending on the capitalization of the next line.

Both bugs are one-line fixes on their own. The reason to refactor instead is that the current shape makes the next bug of the same class easy to write, and the pre-existing flattening defects (blockquote masking, bullet lists, setext headings) cannot be fixed inside a function whose contract is "collapse everything except line 0".

## Proposed shape

Three small pure functions, each unit-testable, with a hard rule that no size constant is reachable from the first two.

```
[...]

Issue #288, opened 2026-09-13
Coordinate #178 preprocessing repair and cross-family review


conorbronsdon:
**Completed:** #290 merged as `940a8812647ada70a97eb16b24e879dfa1512422`. The reviewed head was `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`; the Claude final-head gate is complete. #178 is closed as superseded, #179 as fixed, and #289 as implemented. Non-blocking follow-ups remain in #293 and the separate detector issue #291. [Landing record](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5656454632).

Track the remaining corpus-preprocessing repair from #178 and coordinate independent model-family reviews before it lands. Opened by Codex at Conor's request.

The policy, Stock reaction framing name, stable emotional-flatline API type, and score/highlight behavior already landed through #177 and #287. This work is about the reliability of measurement preprocessing. Preserve Francisco Quinteros (@JavierQuinan)'s #180 work and attribution.

Starting references: main after #287 is fabd62d9c8785dd0edda35201359bcc635b7d3de; the reviewed #178 head is a37404a9545e8b766740737520f76725cf471d65. Refresh both before implementation.

## Coordination

- Implementation session: this Codex session, authorized by Conor's “Proceed to implementation.” Branch `maint/fp-preprocessing-repair`, base `fabd62d9c8785dd0edda35201359bcc635b7d3de`. [Claim and accepted contracts](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5655973504).
- Current Codex session: implementation complete with GPT-5.6 Sol subagents for preprocessing, corpus/comparison, and independent review; final candidate #290. Earlier OpenCode reviews apply only to #178; fresh successor reviews were completed and are recorded below.
- GPT reviewer: GPT-5.6 Sol (medium) reviewed final #290 SHA `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`; no remaining blockers. [Final review and evidence](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5656193515).
[...]

PR #290, opened 2026-09-13
Repair corpus preprocessing and make measurement decisions auditable

### LINKED ISSUE #179: Prevent heading attachment from dropping an oversized measurement unit

conorbronsdon:
Corpus measurement flattened Markdown structure, hiding list and heading findings and preventing quotation masking. The remaining preprocessing work in #178 also deleted content in document mode and could discard eligible paragraphs at the 400-word boundary.

This successor starts from current main and separates structural classification, prose joining, and paragraph selection. Document preparation preserves ordered tokens. Paragraph preparation retains eligible bodies and headings, keeps fenced regions atomic, and preserves detector-visible list and quotation boundaries.

Measurement verifies cached source hashes and records every selected or skipped unit with original UTF-16 spans, input/detector word counts, structural metadata, score/categories, and a reason for exclusions. `--dump-units PATH` writes those records without source text. The new comparison command runs legacy and repaired preparation against the same detector and verified corpus, including zero-count categories and source/register comparisons.

Validation: all 18 npm test suites and all four GitHub check runs pass on `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`. Independent GPT-5.6 Sol review and static MiMo/Ling/Nemotron amendment reviews found no remaining blockers. The separate Claude final-head review recommends merge after independently verifying all 38 corpus hashes, reproducing the published comparison artifacts byte-for-byte, and rerunning 60,000 document/paragraph property cases with zero counterexamples. [Claude's final review](https://github.com/conorbronsdon/avoid-ai-writing/pull/290#pullrequestreview-5192417046).

Non-blocking follow-ups are tracked in #293 and should be addressed before the next measurement is cited. They include mid-paragraph ordinal classification, comparison pairing for attached headings, and tests that pin the legacy/current delta. Headline corpus counts and rates reproduce; the comparison impact buckets currently misclassify some heading attachments.

Candidate: `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`.

[Final amended review, corpus results, and reproducible evidence](https://github.com/conorbronsdon/avoid-ai-writing/issues/288#issuecomment-5656193515). A separate detector boundary follow-up exposed by this run is #291.

Thanks to @JavierQuinan (Francisco Quinteros) for the original boundary fix and regression in #180. His work is carried forward with a Co-authored-by trailer and expanded coverage.

Fixes #179. Implements #289 and the remaining preprocessing scope of #178; coordination and final review gates remain in #288. The detector, corpus selection/hashes, release versions, and already-landed policy are unchanged.


conorbronsdon:
I inspected the later automated review before merging. Two findings reproduce and are being repaired: tab-indented fences accepted by the detector could be split during paragraph preparation, and the default measurement path verified one source read then parsed a second read. The patch now preserves those fences and parses the exact verified source snapshot. It also adds a detector-backed fence regression and a generated exact-once source-span coverage matrix. I will publish the amended SHA and refresh the affected reviews/evidence; the previous final report remains a record of f1dfd940, not approval of the pending amendment.

The proposed old JSONL field aliases and exact exported pass names refer to the original #289 pseudocode, which the reviewed implementation contracts superseded: inputWords/detectorWords avoid the ambiguous words field, kinds and colon-inferred provenance are explicit, and internal classification/normalization are separate from paragraph-only selection. I am retaining that API rather than adding unused aliases or a second public surface.

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+    .map((decision) => decision.text);
+}
+
+module.exports = {
+  MIN_WORDS,
+  MAX_WORDS,
+  prepareUnits,
+  normalizeUnit,
+  splitUnits,
+  unitsForText,
```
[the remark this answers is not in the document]
conorbronsdon: The implementation keeps classify(), normalizedRange(), and paragraph selection as separate pure internal responsibilities; document preparation never invokes the paragraph selection/size-limit path. The original #289 function names and Block sketch were refined during design review because blind blank splitting breaks fences and source-span coverage. The agreed public API is prepareUnits plus compatibility wrappers. Exposing additional parser internals is not needed to establish that separation or the tested invariants; retaining the reviewed API.

scripts/fp-measure.js
```diff
@@ -76,49 +72,143 @@ function rocAuc(posScores, negScores) {
...
+      if (!['human', 'machine'].includes(cls)) throw new Error('Invalid corpus class: ' + cls);
+      const rowSourceHash = sha256(row.text);
+      rowIdentities.push({ doc: doc.id, rowId: row.id ?? rowIndex, rowIndex, rowSourceHash });
+      let index = 0;
+      for (const decision of prepare(row.text, unit).decisions) {
+        const { text, ...selection } = decision;
+        const record = {
+          recordKind: 'unit', doc: doc.id, rowId: row.id ?? rowIndex, rowIndex, cls,
+          register: row.register || doc.register || 'unknown', model: row.model || null,
+          rowSourceHash, ...selection,
```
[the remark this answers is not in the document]
conorbronsdon: The reviewed schema deliberately replaced ambiguous words with inputWords and detectorWords, uses kinds for structural provenance, and labels inferred headings as headingKind: 'colon-inferred'. There was no pre-existing --dump-units consumer contract to preserve. Every selected/skipped decision is included; per-source inferred-heading counts are derived directly from those records and reported in #288 (68 paragraph decisions, 31 selected/37 skipped on the verified corpus). The original proposal's field names were superseded by this versioned schema and its documentation.

scripts/fp-preprocess.test.js
```diff
@@ -0,0 +1,438 @@
...
+  assert.deepEqual(attached[0].spans, [
+    { start: 0, end: 'Context:'.length },
+    { start: 'Context:\r\n'.length, end: attachedInput.length },
+  ]);
+  assert.equal(attached[0].text, `Context:\n${body399}`);
+
+  const body400 = words(400);
+  const detachedInput = `Context:\r\n${body400}`;
+  const detached = prepareUnits(detachedInput).decisions;
+  assert.equal(detached.length, 2);
```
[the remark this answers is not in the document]
conorbronsdon: Addressed in 86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6. Added a deterministic mixed-structure exact-once matrix across LF, CRLF, and lone CR, independently specified original spans, heading/body 399/400 boundaries, selected counts, structural kinds, unique sentinels, and ordered token equality. Existing tests cover floor/ceiling and rejected-unit cases. The affected suite and full 18-suite run pass.

scripts/fp-preprocess.test.js
```diff
@@ -0,0 +1,438 @@
...
+    '~~~',
+    '```',
+    '```` trailing text',
+    words(25, 'more'),
+    ticks,
+  ].join('\r\n');
+  const result = prepareUnits(input);
+  assert.equal(result.decisions.length, 1);
+  assert.equal(result.decisions[0].status, 'selected');
+  assert.deepEqual(result.decisions[0].kinds, ['fenced-code']);
```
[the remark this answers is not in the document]
conorbronsdon: Addressed in 86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6. The eligible tab-indented fence fixture contains an internal heading, five short noun-phrase bullets, prose, and blank lines. It asserts atomic preparation, exact retained layout, and no title-case-header or bullet-np-list findings through the real detector. The assertion is scoped to structural findings because the unchanged detector intentionally still checks some vocabulary inside code.

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+ * Markdown structure is kept where the detector can observe it. Only ordinary
+ * hard-wrapped prose is folded. Paragraph eligibility is deliberately a
+ * separate final step so skipped candidates remain available for diagnostics.
+ */
+
+const MIN_WORDS = 50;
+const MAX_WORDS = 400;
+
+const ATX = /^ {0,3}#{1,6}(?:[ \t]+|$)/;
+const FENCE = /^ {0,3}(`{3,}|~{3,})(.*)$/;
```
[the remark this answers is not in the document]
conorbronsdon: Confirmed and fixed in 86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6. Before the fix, the tab-fence reproduction yielded a skipped opener plus an unwrapped selected body with both title-case-header and bullet-np-list. Openers and closers now match the detector's [ \t]{0,3} indentation grammar and retain same-marker/minimum-length closure. The entire example is one fenced-code decision and those findings disappear. Regression, independent Sol amendment review, and CI pass.

scripts/fp-measure.js
```diff
@@ -76,49 +72,143 @@ function rocAuc(posScores, negScores) {
...
+    if (textFor) {
+      const text = textFor(doc);
+      unavailable = text === null;
+      if (!unavailable) {
+        source.sha256 = sha256(text);
+        if (!doc.sha256 || source.sha256 !== doc.sha256) throw new Error('Corpus hash mismatch or missing hash: ' + doc.id + '; run corpus.js verify');
+        source.status = 'verified';
+      }
+    }
+    const rows = unavailable ? null : rowsFor(doc);
```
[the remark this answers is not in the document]
conorbronsdon: Confirmed and fixed in 86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6. Added corpus.rowsFromText(doc, text); loadRows delegates to it for existing callers, while default measure() passes the already verified source snapshot directly, avoiding the second read. The regression changes a local source after its 60-word snapshot is loaded and proves measurement still scores/hashes that snapshot rather than the replacement; dataset parsing is covered too. Explicit injected row loaders remain a caller-owned test seam. All 38 corpus decisions/metrics are unchanged on regeneration, and all 18 suites pass.

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+ * hard-wrapped prose is folded. Paragraph eligibility is deliberately a
+ * separate final step so skipped candidates remain available for diagnostics.
+ */
+
+const MIN_WORDS = 50;
+const MAX_WORDS = 400;
+
+const ATX = /^ {0,3}#{1,6}(?:[ \t]+|$)/;
+const FENCE = /^[ \t]{0,3}(`{3,}|~{3,})(.*)$/;
+const LIST = /^[ \t]*(?:[-+*•]|\d+[.)])[ \t]+/;
```
conorbronsdon, line 16: **Blocking: an ordinal mid-paragraph turns the rest of the block into `list` and disables prose joining.** `LIST` matches `\d+[.)]` plus space anywhere, and the run loop at 121 to 141 then marks every following non-blank line `list`, so hard wraps stop being joined for the remainder of the paragraph. Years at a line start are common in the 1788 to 1907 sources.

```js
prepareUnits('The decisive year was\n1859. The publication of that work changed the temper of the debate,\nand no one who has read it can afford to pass over the evidence.', 'document')
// kinds ["prose","list"]; output keeps both newlines instead of joining. Control "in 1859. The publication" joins.
```

CommonMark only lets an ordered item interrupt a paragraph when the marker is `1.`; a `1859.` line after a prose line is paragraph continuation. Suggested rule: start a list run only when the previous line is blank or structural, or when the marker is a bullet or `1.`; and once inside an item, join lazy continuation lines to the item text rather than marking them `list`. Add the `1859.` case as a regression.

---

scripts/fp-compare.js
```diff
@@ -0,0 +1,442 @@
...
+    },
+  };
+  const sources = [...new Set([...Object.keys(legacy.bySource || {}), ...Object.keys(current.bySource || {})])].sort();
+  result.bySource = Object.fromEntries(sources.map((source) => [source,
+    compareAccounting(legacy.bySource?.[source] || {}, current.bySource?.[source] || {}),
+  ]));
+  return result;
+}
+
+function identity(record) {
```
conorbronsdon, line 137: **Blocking: the comparison cannot pair a unit whose heading was attached, which is the PR's central change.** `identity()` includes `spans`. Attachment turns one span into two, so the legacy record (body span only) and the current record (heading span plus body span) never match, and the change is reported as one `added` plus two `removed` rather than one `modified` with `impact: 'detector'`.

```js
// row text: '## Migration Notes\n\n' + 60 words; compare({ units: ['paragraph'] })
// counts {"added":1,"removed":2,"modified":0}; byImpact {"population":3,"detector":0}
```

So `byImpact.detector` is zero for exactly the units whose scored text changed. Match on the body span (the last span, or any overlapping span) and treat the heading span as a field change, so attachment shows up as `modified` with the score and type deltas visible.

---

scripts/fp-preprocess.test.js
```diff
@@ -0,0 +1,533 @@
...
+  const indentedQuote = `${words(20)}\n    > Quoted one ${words(14, 'quote')}\n\t> Quoted two ${words(14, 'more')}`;
+  const preparedIndentedQuote = prepareUnits(indentedQuote).decisions;
+  assert.equal(preparedIndentedQuote.length, 1);
+  assert.equal(preparedIndentedQuote[0].status, 'selected');
+  assert.deepEqual(preparedIndentedQuote[0].kinds, ['prose', 'blockquote']);
+  assert.equal(preparedIndentedQuote[0].text, indentedQuote);
+  assert.equal(
+    AIDetector.analyzeText(preparedIndentedQuote[0].text).stats.wordCount,
+    AIDetector.analyzeText(indentedQuote).stats.wordCount,
+  );
```
conorbronsdon, line 344: **Blocking, small: no test pins the legacy-versus-current detector delta the PR exists to produce.** Line 340 asserts the prepared text equals the raw input, and 341 to 344 then compare `analyzeText(x)` with `analyzeText(x)`. The same holds for the Unicode-bullet, indented-bullet, and five-quote-line fixtures earlier in the file: preparation leaves them byte-identical, so every detector-visible assertion passes with a no-op preprocessor.

Four assertions using `legacyPrepareUnits` from `fp-measure.js` would close this. Measured on this head:

| fixture | legacy | current |
|---|---|---|
| five `> ` lines with flaggable text | score 4, 2 issues | score 0, 0 issues |
| five noun-phrase bullets after prose | `[]` | `["bullet-np-list"]` |
| setext heading plus body | `[]` | `["title-case-header"]` |
| fence containing bullets and `#` | `[]` | `[]` |

---

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+      const oneBlankLine = next === i + 1;
+      const structuralContinuation = oneBlankLine && next < lines.length && (
+        (sawList && classified.kinds[next].includes('list'))
+        || (sawQuote && classified.kinds[next].includes('blockquote'))
+        || (sawIndented && classified.kinds[next].includes('indented-code'))
+        || (sawProse && !sawList && classified.kinds[next].includes('list'))
+      );
+      const prospectiveEnd = structuralContinuation ? nextBlockEnd(lines, classified, next) : next;
+      const mergeFits = structuralContinuation
+        && wordCount(normalizedRange(lines, classified, start, prospectiveEnd)) <= MAX_WORDS;
```
conorbronsdon, line 303: `MAX_WORDS` reaches block segmentation here, so structure depends on size and the record cannot show it. With five two-word bullets after a blank line: 385 prose words gives one 400-word unit with `kinds ["prose","list"]`; 386 prose words gives a 386-word prose unit plus a 15-word list skipped `below-min`, and the bullets leave the measurement. Both records look like ordinary blocks. Not blocking, but either move the fit check into selection (merge unconditionally, let eligibility decide) or record `mergedContinuation: true` so the decision is auditable. Filing this as a follow-up is fine.

---

scripts/fp-measure.js
```diff
@@ -76,49 +72,146 @@ function rocAuc(posScores, negScores) {
...
+    let verifiedText;
+    if (textFor) {
+      verifiedText = textFor(doc);
+      unavailable = verifiedText === null;
+      if (!unavailable) {
+        source.sha256 = sha256(verifiedText);
+        if (!doc.sha256 || source.sha256 !== doc.sha256) throw new Error('Corpus hash mismatch or missing hash: ' + doc.id + '; run corpus.js verify');
+        source.status = 'verified';
+      }
+    }
```
conorbronsdon, line 157: The verified bytes are not the scored bytes. `textFor(doc)` reads and hashes the file once, then `rowsFor(doc)` goes through `loadRows`, which calls `loadText` and reads it again (two reads per source per `measure`, eight for a two-mode `compare`). Derive the rows from the `text` you already verified: for dataset sources split the JSONL here, otherwise wrap it as the single row, and only fall back to `rowsFor` when a loader is injected. Also worth noting in the docs: passing `loadRows` without `loadText` skips verification entirely and records `status: 'injected'`, which is the shape every test uses. Low severity, cheap fix.

---

scripts/fp-measure.js
```diff
@@ -76,49 +72,146 @@ function rocAuc(posScores, negScores) {
...
+          }
+        }
+        records.push(record);
       }
     }
   }
-  return { units, skipped, unit };
+  const metadata = {
+    schemaVersion: 1, unit, preprocess, revision: revision(),
+    legacyReference: 'fabd62d9c8785dd0edda35201359bcc635b7d3de',
```
conorbronsdon, line 207: `preprocessorHash` is the same value for both paths, so `preprocessorHashes.legacy` and `.current` in the comparison metadata are always equal (verified: both `eaec1137…`). Hash the function that actually ran, for example `String(prepare)` plus `fp-preprocess.js` for current and `String(legacyPrepareUnits)` for legacy, or drop the pair and keep one code fingerprint plus the `preprocess` label.

---

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+const ATX = /^ {0,3}#{1,6}(?:[ \t]+|$)/;
+const FENCE = /^[ \t]{0,3}(`{3,}|~{3,})(.*)$/;
+const LIST = /^[ \t]*(?:[-+*•]|\d+[.)])[ \t]+/;
+const LIST_CONTINUATION = /^(?: {2,}|\t)\S/;
+const QUOTE = /^[ \t]*>/;
+const INDENTED = /^(?: {4}|\t)/;
+const SETEXT = /^ {0,3}(?:=+|-+)[ \t]*$/;
+const THEMATIC = /^ {0,3}(?:(?:\*[ \t]*){3,}|(?:-[ \t]*){3,}|(?:_[ \t]*){3,})$/;
+
+function wordCount(text) {
```
conorbronsdon, line 23: Heading markers count as words, so punctuation moves the eligibility boundary. `'## ' + 49 words` is `inputWords: 50` and selected as a paragraph; `'## Context\n' + 398 words` attaches at 400 while `'#### A B\n' + 398 words` is `heading-would-exceed-maximum` at 401. Two tests (166 to 179) pin this as the contract. Either count words after stripping `#` prefixes and setext underlines, or say in the README that markers count. Low.

---

scripts/fp-preprocess.js
```diff
@@ -0,0 +1,479 @@
...
+ * Markdown structure is kept where the detector can observe it. Only ordinary
+ * hard-wrapped prose is folded. Paragraph eligibility is deliberately a
+ * separate final step so skipped candidates remain available for diagnostics.
+ */
+
+const MIN_WORDS = 50;
+const MAX_WORDS = 400;
+
+const ATX = /^ {0,3}#{1,6}(?:[ \t]+|$)/;
+const FENCE = /^[ \t]{0,3}(`{3,}|~{3,})(.*)$/;
```
conorbronsdon, line 15: The detector's `fenceRanges` accepts `[ \t]{0,3}` before a fence marker; this accepts spaces only. A tab-indented fence with a blank line inside is classified `indented-code` plus `list` and split into two atoms, while the detector would treat the whole region as one fence. CommonMark is on your side here (a tab is four columns, so it is indented code), so the disagreement is the detector's. Either match the detector's regex for measurement consistency or note the divergence in `corpus/README.md`. Low.

---

conorbronsdon (commented):
## Claude final-head review for #288, head `f1dfd940559ab8cf3041ec6465ac79a0ad0b9e72`

Recommendation: **changes requested** (GitHub does not allow the formal Request Changes state on the author's own PR, so this is submitted as a comment review).

**Session/client.** Claude Code (web), session `session_01LmCsufeKSGpesBP7w4Xf4j`, coordinating model `claude-fable-5-1`. One independent Claude Opus subagent reviewed isolated checkouts of this head and of base `fabd62d` with no GitHub access and no other reviewer's conclusions; I re-executed every finding below on the same checkouts before posting. Nothing here is inference-only.

**What ran.** `npm test` on the head: 18 suites pass. My own replay of every defect I reported on the #178 head (document-mode heading and clause deletion, `## Context` plus lowercase body, consecutive headings, 399/400/401 boundaries, blockquote masking, `bullet-np-list`, setext, `~~~` inside a backtick fence, 20 versus 21 word colon lines): all fixed on this head. The Opus pass fuzzed 40,000 document-mode inputs for token preservation and 20,000 paragraph-mode inputs for exact-once coverage, span disjointness, and span-to-text agreement, with zero counterexamples. Timing: a 6,000-word hard-wrapped document prepares in about 10 ms; 50k-character runs of spaces, dashes, and `* ` show no backtracking. CI is green on all four check runs. I did not run the corpus measurement (no cache in this environment).

**Verdict: the repair is correct and the invariants hold. Three things should change before merge, all small.**

1. `scripts/fp-preprocess.js:16` A line starting with a number and a period (for example a year, `1859.`) mid-paragraph is classified as a list item and switches off prose joining for the rest of the block. This affects the 1788 to 1907 sources directly.
2. `scripts/fp-compare.js:137` The changed-decision matcher keys on the full span list, so a unit whose heading was attached shows as one `added` plus two `removed`, and `byImpact.detector` reads zero for exactly the units the repair changed.
3. `scripts/fp-preprocess.test.js:340` The detector-visible tests analyze text the preparation leaves byte-identical, and one block compares a result with itself. Four assertions against `legacyPrepareUnits` would pin the delta this PR exists to produce; the values are in the inline comment.

Non-blocking, inline: `MAX_WORDS` reaching block segmentation through the blank-line list merge (unrecorded in the unit record); the source read twice so the verified hash covers different bytes than the scored rows; `preprocessorHashes.legacy` always equal to `.current`; heading markers counting as words at the 50/400 boundaries; space-only fence indentation versus the detector's `[ \t]{0,3}`.

One documentation note that could not be attached inline because the line is unchanged: `corpus/README.md:201`, the `Results (v3.22.0, 2026-07-31)` table, now sits directly under a present-tense description of the repaired pipeline, and `PROOF.md:111-114` quotes the same figures. One sentence saying those numbers came from the legacy path, with a link to the `fp-compare` artifact from the #288 evidence comment, would keep the two from reading as one measurement.

Reconciliation with the Qodo review on this PR: its finding 5 (tab fences) and 6 (double read) reproduce and are covered above at low severity. Its finding 2 asks for the field names from #289 (`words`, `kindsPresent`, `manufacturedHeading`); the record already carries the same information as `inputWords`, `kinds`, and `headingKind: 'colon-inferred'`, so no change is needed there. Its finding 1 (fused passes) is a structure preference; the invariants the split was meant to protect hold under fuzzing, so I would not block on it.

Thanks to @JavierQuinan for the #180 boundary fix, which is carried forward correctly: the 400-word body after a heading is retained on its own and the 399-word body plus heading combine, in both single- and double-newline forms.

---

conorbronsdon (commented):
## Claude final-head review: `86ef5ab3e2f85199f7b2d6cd2231ba9ff91a49b6`

**Recommendation: merge.** The head matches Codex's reviewed commit. Every claim I could test independently held, my full reproduction of the published comparison is byte-identical, and the findings that remain do not change any reported number. They are filed as #293 to land before the next measurement is cited. (GitHub blocks the formal Approve state on the author's own PR, so this is a comment review.)

**Session/client.** Claude Code (web), session `session_01LmCsufeKSGpesBP7w4Xf4j`, model `claude-fable-5-1`. My earlier review on this PR was of `f1dfd940`; this one is of the amended head. The independent Claude Opus fuzzers from that pass were re-run against `86ef5ab3`.

### Tests I ran (executed, not static)

| Claim | Result |
|---|---|
| `npm test` on `86ef5ab3` | 18 suites pass |
| CI on the head | 4 check runs green |
| Corpus: `corpus.js fetch` then `verify` | 38 of 38 sources verified, zero unavailable (archived blog pages need `NODE_USE_ENV_PROXY=1` in this sandbox; the script's own retry got HTTP 403 without it) |
| `fp-compare.js --out` on the verified corpus | `decisions.jsonl` SHA-256 `5e32f6db…` and `summary.json` SHA-256 `580f30d9…`, identical to the published artifacts; `code.sha256` matches the head's six files |
| Comparison table | paragraph human 889 to 891, machine 779 to 792; document human 376 to 376, machine 530 to 529; FPR/TPR at 3, 5, 10 match the #288 report to three decimals |
| Document-mode token preservation | 40,000 generated inputs mixing every line kind, CRLF, lone CR, Unicode bullets, colon lines: zero counterexamples |
| Paragraph coverage | 20,000 generated inputs: every 50 to 400 word block in exactly one span, no overlaps, 401-word blocks recorded `above-max`; matrix of five heading shapes, two separators, 49/50/399/400/401 bodies, upper and lowercase |
| Heading boundaries | `Context:` plus 399 attaches at 400; plus 400 keeps the body alone and records `heading-would-exceed-maximum`; `## Context` plus lowercase body preserved in both modes; consecutive headings attach the nearest one |
| Markdown structure | five `> ` lines stay masked (score 4 to 0 versus legacy); five bullets keep `bullet-np-list`; setext keeps `title-case-header`; `~~~` inside a backtick fence does not close it; tab-indented fence with an internal blank line is now one `fenced-code` decision (fixed in this head) |
| Snapshot rows | `measure()` reads each source once and builds rows from the verified bytes (fixed in this head); the new test mutates the file after the read and proves the snapshot is what gets scored |

### Findings on this head, all reproduced, none blocking

Two of my three earlier blocking items are unchanged here, and I quantified them on the real corpus before deciding they should not hold the merge:

1. **Mid-paragraph ordinals disable prose joining.** `1774.` in Federalist and `1758.` in Franklin are the only two hits across 906 rows; 16 lines stay unjoined; score and types are unchanged for both rows in both modes.
2. **Comparison pairing of attached headings.** 62 of 70 attachments are reported as `added` plus `removed` rather than `modified`; 8 of those changed score, so `byImpact.detector` undercounts. The headline counts and rates come from the runs, not the change set, and are unaffected.
3. **Legacy-delta tests.** The detector-visible tests still analyze text the preparation leaves byte-identical; the delta is demonstrated by the corpus artifact, not by the suite.

Smaller items (segmentation reaching `MAX_WORDS`, equal preprocessor hashes, markers counted as words, the results table in `corpus/README.md` sitting under the new pipeline description) are listed in #293 with reproductions.

### Reconciliation

Codex's amended report is accurate on every point I could check: suite count, CI, 38 verified sources, unchanged decision stream, and the changed-example explanations. Qodo's tab-fence and double-read findings are fixed. The Ling and Nemotron objections Codex rejected were rejected correctly; I re-executed the `--unit=VALUE` contract and the equal-span normalization case. Francisco's #180 boundary behavior is carried forward intact.

---

PR #300, opened 2026-09-14
fix: allow pre-commit args to override gate defaults

### LINKED ISSUE #243: Pre-commit hook crashes whenever a project overrides options via args
`.pre-commit-hooks.yaml` ends its `entry` with a bare `--`:

```yaml
entry: avoid-ai-writing-gate --context technical --source-mode rendered-markdown --threshold 6 --
```

`bin/avoid-ai-writing-gate.js` treats the first `--` as "everything after is a filename". pre-commit builds the command as the hook's `entry`, then the project's `args:`, then the matched filenames. So any `args:` override lands after the `--` and is read as a file path. Reproduced end to end with the real `pre-commit` tool against this repo:

```yaml
hooks:
  - id: avoid-ai-writing
    args: [--threshold, "0"]
```

```
$ pre-commit run --files sample.md
avoid-ai-writing-gate: cannot read --threshold: ENOENT: no such file or directory, open '--threshold'
```

`args:` is the standard, documented way to customise a pre-commit hook, and the README itself presents `threshold: "0"` as a legitimate project choice. The default configuration (no override) works.

### Decision needed

Drop the trailing `--` and make flags repeatable so a later `--threshold` overrides an earlier one (simplest for users; needs the parser to accept repeats), or split into hook variants, or document that this hook cannot be customised via `args:` and must be re-declared as a local hook.

### Scope

`.pre-commit-hooks.yaml`, possibly the argument parser in `bin/avoid-ai-writing-gate.js`, and a regression test that drives the real `pre-commit` tool or simulates its argument order exactly. Document the outcome in the README's pre-commit section.

### Acceptance criteria

- A config with `args: [--threshold, "0"]` applies the override instead of crashing.
- The default (no override) behaves as today.
- A test covers pre-commit's argument order.

Add an `## [Unreleased]` changelog entry: installation and configuration behavior.

kevin-lozada-santos:
Pre-commit appends user `args` after the hook entry. The entry's hard-coded `--` therefore made `--threshold` (and other overrides) look like filenames. Move the separator to default `args: ["--"]`, preserve entry defaults, and document ending custom args with `--` for dash-prefixed filenames.

Adds regression coverage against the actual manifest for defaults, strict/permissive thresholds, source-mode override, context/source-mode validation and dash-prefixed files. Includes the requested Unreleased entry. No parser change.

Validation:
- The new strict-hook regression failed before the fix (`cannot read --threshold`, exit 2 instead of 1), then passed.
- `node bin/avoid-ai-writing-gate.test.js` passes.
- Real pre-commit 4.6.2 invocation in a disposable Git repository: default and permissive configurations passed; strict configuration correctly rejected `-draft.md`. This uses the manifest entry/args with the existing Node executable via a local system hook, so it tests pre-commit argument assembly but not installation of the remote Node hook environment.
- `npm test`: 18 test scripts pass; `scripts/rewrite-eval-opencode.test.js` fails on native Windows with `spawnSync fake-opencode.js EFTYPE`. The identical test fails on unchanged base `366afe4` with the same error.
- `git diff --check` passes.

Fixes #243. Implements the configuration-only approach approved in the issue.


conorbronsdon (approved):
Reviewed at 89f43cc9c94de4e2f1cd9ff1a9633588b814f440. The hook now preserves defaults, permits later args overrides, and protects dash-prefixed filenames through the documented separator. Independent Claude and isolated zero-cost OpenCode reviews found no runtime defect. Local targeted gate tests passed; the bare threshold override also parsed correctly. GitHub checks are green, and the only local full-suite failure is the known Windows-only EFTYPE case reproduced unchanged on the base commit.

kevin-lozada-santos:
I’ve really enjoyed contributing here. I’m currently looking for paid software/dev-tools work. If your team is hiring or needs contract help, I’d love to talk.

conorbronsdon:
Sorry @kevin-lozada-santos I'm not currently hiring for any roles, but appreciate the thought and thank you for the contribution!

PR #305, opened 2026-09-14
fix: preserve query params when stripping trackers

### LINKED ISSUE #210: Strip AI tracking parameters without breaking URLs that carry other query params

Item 3 from #143, never fixed. `normalizeUrl()` in `detector/validate.js` removes `?utm_source=chatgpt.com` but leaves the `&` that followed it, so the normalised original and rewrite no longer match.

```
$ node -e "
const {validate}=require('./detector/validate.js'); const o={skipResidual:true};
const a='See [post](https://example.com/a?utm_source=chatgpt.com&ref=home) for details of the thing.';
const b='See [post](https://example.com/a?ref=home) for details of the thing.';
console.log(JSON.stringify(validate(a,b,o).errors));"
[{"code":"url-missing","message":"URL removed or altered: https://example.com/a&ref=home, https://example.com/a&ref=home"}]
```

The single-parameter case passes. This is a false positive on the skill's own documented edit: `SKILL.md` tells the model to strip the AI-referrer parameter, and the carve-out in `validate.js` exists so that edit is never blocked.

### Scope

Rewrite `normalizeUrl()` in `detector/validate.js` to parse and rebuild the query string, or to preserve the leading `?` when the first parameter is removed. Add table-driven fixtures to `detector/validate.test.js` covering `?utm=...&b=1`, `?a=1&utm=...`, `?a=1&utm=...&b=2`, and the existing single-parameter case. Run `bash scripts/sync-plugin-skill.sh` afterwards.

### Acceptance criteria

- Stripping `utm_source=chatgpt.com` from a URL with one, two, or three other parameters produces no `url-missing`, in either parameter order.
- A URL whose non-tracking query string actually changed still raises `url-missing`.
- `referrer=grok.com` behaves identically.
- `npm test` passes and `git diff` is empty after the sync script.

Add an `## [Unreleased]` changelog entry: this changes validator behavior.

LucasLeao18:
## Summary

`normalizeUrl` previously removed matching tracking text in place. When the tracking parameter came first, this could leave a leading `&`, so a URL that still carried useful query parameters was reported as missing.

This change rebuilds the query string after removing exact `utm_source=chatgpt.com` and `referrer=grok.com` parameters. It preserves every other parameter in its original order, retains fragments, and avoids broader URL normalization.

Regression cases cover both tracking parameters at the beginning, middle, and end of a query string, plus a real non-tracking URL mutation to ensure meaningful changes are still detected.

## Checklist

- [x] `npm test` passes (20 suites passed, 0 failed)
- [x] No detector `type` was added
- [x] No judgment-only rule was added
- [x] False positives were considered; only exact tracking parameter names and values are removed
- [x] No factual claims about AI or human writing were added
- [x] Added an Unreleased `CHANGELOG.md` entry for the user-facing fix
- [x] This PR does not prepare a release
- [x] No detection category was added or removed
- [x] `scripts/sync-plugin-skill.sh` was run twice and was idempotent
- [x] `git diff --check` passes

Closes https://github.com/conorbronsdon/avoid-ai-writing/issues/210


LucasLeao18:
Addressed the review findings in 2287b7c: added the missing Grok middle/sole cases, normalized leading/trailing/doubled empty query fields only when an exact tracker is present, and preserved sentence punctuation captured after a terminal tracker. The focused validator suite and the full 20-suite npm test run pass, and the generated validator copies were resynchronized.

conorbronsdon (approved):
Approved at exact head b9eb7729767f08c2adcf6b442b5496f15abded86. The asymmetric empty-field and terminal-question-mark regressions were reproduced, repaired, and retested. Focused preservation tests pass; full suite is 20/20; self-scan, sync, diff checks, and all live GitHub checks pass; generated validator copies are identical. Claude Opus 5 and Astra found no remaining actionable issue. Hermes' empty-field concern is non-actionable because normalization deliberately canonicalizes empty fields symmetrically while preserving all meaningful parameters.

PR #310, opened 2026-09-14
fix(detector): make acknowledgment loops judgment-only

### LINKED ISSUE #239: Narrow or retire acknowledgment-loop: it fires on ordinary question framing
`ACKNOWLEDGMENT_LOOPS` in `detector/patterns.js` has three patterns: `you're asking about`, `the question of whether`, and `to answer your question`.

```
$ node -e "
for (const t of [
 'The question of whether the effect persists after controlling for income is still open, and the two replications disagree with each other.',
 'To answer your question from Tuesday: the invoice went out on the 3rd and the payment cleared last week, so nothing is outstanding on our side.',
 \"You're asking about the retry limit. It is five by default and configurable with RETRY_MAX, though we do not recommend raising it past ten.\"])
 console.log(JSON.stringify(D.analyzeText(t,{contextMode:'technical'}).issues.map(i=>i.type+':'+i.text)));"
["acknowledgment-loop:The question of whether"]
["acknowledgment-loop:To answer your question"]
["acknowledgment-loop:You're asking about"]
```

"The question of whether" is standard academic and analytical English. The other two are how people open a reply. The tell the rule targets is a chatbot restating the prompt before answering, which requires the restatement to add nothing, and that is not detectable from these phrases alone. The category has no fixtures, and the affected registers (email, support, docs) are exactly the ones the human control corpus lacks (#85), so `fp-measure.js` cannot see this.

### Decision needed

Narrow it (drop `the question of whether`; require the other two to be document-initial or immediately followed by a restatement of the same noun phrase) or move the category to the Skill-only section of `CATEGORIES.md` as judgment-only, with the reasoning recorded like the existing "wall-of-text replies" and "narrated candor" entries.

### Scope

`detector/patterns.js`, fixtures, `detector/CATEGORIES.md`, and `references/patterns.md` if the documented shape changes. Run both sync scripts.

### Acceptance criteria

- The three sentences above stay clean.
- A genuine loop still fires, or the category is documented in the Skill-only section with its reasoning.
- A must-fire and a must-not-fire fixture land for whatever remains.

dchaudhari7177:
## Summary

Fixes #239. This takes the second option: `acknowledgment-loop` is now judgment-only.

Narrowing didn't hold up. The issue suggested requiring the two reply openers to be document-initial, but both false positives in the issue already are: "To answer your question from Tuesday: ..." and "You're asking about the retry limit. ..." each open their document. The remaining condition, a restatement of the same noun phrase, needs to know whether the restatement adds anything, and the rule already sits in that territory. "The question of whether" gets dropped either way.

Changes:

- `detector/patterns.js`: removed `ACKNOWLEDGMENT_LOOPS`, its weight, label and match call, and left a NOTE pointing at §C, as the wall-of-text entry does.
- `detector/CATEGORIES.md`: removed the §A row and added the rule to §C with the reason. The type total goes from 54 to 53 here and in `README.md`.
- `references/patterns.md`: dropped "The question of whether" from the examples and added the deletion test, a reply carve-out, an analytical-framing note, and the judgment-only note. The `###` heading stays, so the pattern count is unchanged (`check-pattern-count.sh` still reports 74).
- `detector/patterns.test.js`: a must-not-fire fixture for the three sentences from the issue, under the default and `technical` contexts. With `main`'s `patterns.js` it fails on all three. No must-fire fixture, since the type no longer exists.
- `CHANGELOG.md`: an Unreleased entry under Changed.
- Generated copies from `sync-plugin-skill.sh` and `sync-cursor-rules.sh`.

## Checklist

- [x] `npm test` passes (engine fixtures + `CATEGORIES.md` contract check): 20 passed, 0 failed
- [ ] If I added a detector `type`: n/a, this removes one
- [x] If I added a judgment-only rule: it's listed under "Skill-only" in `detector/CATEGORIES.md`
- [x] Any factual claim about how AI or humans write cites a source: no new claims. The false positives are the three sentences from the issue.
- [x] The prose I added passes the skill's own audit: `npm run self-scan:check` passes, CHANGELOG.md exempt score 34 of 40
- [x] I added an Unreleased `CHANGELOG.md` entry for a user-facing change
- [ ] If this PR prepares a release: n/a
- [x] If I added or removed a detection `###` in `references/patterns.md`: none removed, and `check-pattern-count.sh` passes


conorbronsdon (approved):
Approved at a7deb1dea2f576ac0be1736d5ed43159574d7307. I merged current main into the contributor branch and resolved the sole CHANGELOG conflict by preserving all Unreleased entries. The full 20-suite local test run passed, all refreshed CI checks are green, Qodo reports no issues, and independent Claude Sonnet 5 plus isolated Muse Spark exact-head reviews found no remaining findings. The later main changes merge cleanly; I verified the resulting CHANGELOG retains both sets of entries. Thanks @dchaudhari7177.

conorbronsdon:
Final branch-protection refresh verified at be52079f1dcab52f393e1e47cf99c2dcf57ce028: 20/20 local suites pass, all three GitHub checks are green, and Claude Sonnet 5 plus isolated Muse Spark re-reviews found no issues. The PR delta is byte-identical in substance to the previously reviewed change; the refreshed base merged without conflict.

Issue #255, opened 2026-09-12
Add SECURITY.md, a code of conduct, and CODEOWNERS


conorbronsdon:
```
$ find . -iname "CODE_OF_CONDUCT*" -o -iname "SECURITY*" -o -iname "CODEOWNERS*"
(no results)
```

`PRIVACY.md` and `TERMS.md` cover the OpenAI plugin's data handling, not vulnerability disclosure. `SUPPORT.md` points only at the public issue tracker, which is the wrong venue for an undisclosed vulnerability in a tool that runs in CI and pre-commit on other people's repositories. GitHub's community profile also flags the missing code of conduct.

### What the maintainer needs to supply

- A private disclosure channel for `SECURITY.md` (an email address or GitHub's private vulnerability reporting, which needs enabling in repo settings).
- Which code of conduct to adopt and the enforcement contact.

With those two answers, a contributor can write all three files.

### Scope

`SECURITY.md` with the disclosure channel, supported versions, and expected response window. `CODE_OF_CONDUCT.md` (Contributor Covenant is the common choice) with a real contact. `.github/CODEOWNERS` with at least a default owner so reviews auto-request. Link `SECURITY.md` from `SUPPORT.md`.

### Acceptance criteria

- All three files exist with real, working contact information.
- GitHub's community profile shows them as present.
- `SUPPORT.md` points security reports to the private channel.

Docs only: no changelog entry.

conorbronsdon:
Closing as not planned for the remaining scope. SECURITY.md, private vulnerability reporting, CODEOWNERS, and the security-reporting link from SUPPORT.md are already in place. We are intentionally deferring adoption of a Code of Conduct and its private reporting/enforcement process; PR #276 was closed on that basis. The original acceptance criteria are therefore only partially completed. The conduct-policy work can be revisited when we are ready to maintain that process.

PR #276, opened 2026-09-13
docs: prepare code of conduct pending private reporting contact

sharadvc:
## Summary

Prepares Contributor Covenant 2.1 as a draft pending a maintainer-designated private conduct-reporting contact. The proposal explicitly distinguishes GitHub Support's abuse-reporting process from a private route to project maintainers.

Current main already contains SECURITY.md, SUPPORT.md's security-reporting link, and CODEOWNERS. The branch now incorporates main and retains those files exactly, resolving the stale conflicts and removing the earlier proposed security-response deadlines. The original contribution and commit history are preserved.

Private vulnerability reporting is enabled. The remaining blocker is the private conduct-reporting contact and policy-adoption decision. This PR must remain open and unmerged until those are settled.

Related to #255; this draft does not close it.

## Validation

- [x] `npm test`: 20 suites passed, zero failed.
- [x] `npm run self-scan:check` passed.
- [x] Scoped `git diff --check` passed; no unresolved merge conflicts.
- [x] SECURITY.md, SUPPORT.md, and CODEOWNERS match current main.
- [x] Contributor-guidance-only draft; changelog-exempt under CONTRIBUTING.md.

## Before merge

- [ ] Designate and publish a working private conduct-reporting contact.
- [ ] Confirm policy adoption and the proposed enforcement/response commitments.
- [ ] Replace the missing-contact paragraph (currently lines 64-66), remove the draft banner (lines 3-4), and revise conditional commitments (lines 73-77) when adoption is approved.
- [ ] Review the resulting exact commit again before merging.

Issue #217, opened 2026-09-12
List the three unmapped catalog rules in CATEGORIES.md


conorbronsdon:
Reserved for someone making their first contribution to this repository. Please comment before starting and follow the [one-issue newcomer guidance](https://github.com/conorbronsdon/avoid-ai-writing/blob/main/CONTRIBUTING.md#choosing-an-issue).

`detector/CATEGORIES.md` says its three sections "account for every entry on both sides" and that the Skill-only section exists so an absence is never mistaken for a coverage gap. Three detection headings in `references/patterns.md` appear in none of them:

- `### List-label periods` (line 350)
- `### Speculative gap-filling` (line 369)
- `### Infomercial engagement hooks` (line 392)

```
$ grep -n "List-label periods\|Speculative gap-filling\|Infomercial engagement" detector/CATEGORIES.md
(no results)
```

None has a detector type, and none fires on its own example text, so they belong in the Skill-only section with a reason. Also, `Manufactured punchlines and staccato drama` is listed under a different wording, and the two writer-side tests are not accounted for.

### Scope

Add three bullets to the Skill-only section of `detector/CATEGORIES.md` with a one-line reason each, align the punchlines wording with the catalog heading, and note the writer-side tests as intentionally out of scope. Run `bash scripts/sync-plugin-skill.sh` because the file is copied into both bundles.

### Acceptance criteria

- All three headings appear in `CATEGORIES.md` with a reason.
- A diff of `###` headings in `references/patterns.md` against names in `CATEGORIES.md` returns nothing, or only the writer-side tests, which the file explicitly scopes out.
- `npm test` passes and `git diff` is empty after the sync script.

Docs only: no changelog entry.

Issue #207, opened 2026-09-12
Accept all four --context values in the scoring CLI


conorbronsdon:
Reserved for someone making their first contribution to this repository. Please comment before starting and follow the [one-issue newcomer guidance](https://github.com/conorbronsdon/avoid-ai-writing/blob/main/CONTRIBUTING.md#choosing-an-issue).

The two binaries in the npm package disagree about the same option. `bin/avoid-ai-writing.js:31` declares `CONTEXTS = ["general", "technical"]`, while `bin/avoid-ai-writing-gate.js:27` and the engine (`VALID_CONTEXT_MODES` in `detector/patterns.js`) accept `general`, `technical`, `marketing`, and `personal`. `detector/README.md` documents all four.

```
$ node bin/avoid-ai-writing.js --context marketing README.md
avoid-ai-writing: invalid --context value: marketing
exit=2
```

The scoring CLI's `--help` also advertises only `<general|technical>`.

### Scope

Widen `CONTEXTS` and the `USAGE` text in `bin/avoid-ai-writing.js` to match the gate CLI, and add cases to `bin/avoid-ai-writing.test.js`. No engine change.

### Acceptance criteria

- `--context marketing` and `--context personal` exit 0 and report the requested mode in `stats.contextMode`.
- `--context nonsense` still exits 2.
- `--help` lists the same value set in both binaries.
- A CLI test covers each accepted value and one rejected value.
- `npm test` passes.

Add an `## [Unreleased]` changelog entry: this changes a public CLI option contract.

mateuscmtropical:
Hi, I've opened a PR for this: #327

PR #318, opened 2026-09-15
docs: add CODEOWNERS (#255)

Diogo-Damasceno:
Adds a CODEOWNERS file assigning the default owner. Addresses part of #255 (SECURITY.md is already present in the repo).


conorbronsdon:
Thanks for taking this on, @Diogo-Damasceno — the wildcard rule is the right basic shape for assigning a default owner.

The same need was addressed just before this PR opened: #287 carried the CODEOWNERS contribution from #276 into main, with explicit contributor credit. Main now has a more complete `.github/CODEOWNERS` containing the wildcard owner plus ownership for the canonical skill and detector paths. GitHub checks `.github/` before the repository root and uses the first CODEOWNERS file it finds, so this root-level copy would be ignored rather than adding anything.

I’m closing this as already addressed, not because there was a problem with your implementation. Thank you for spotting the governance gap and submitting the fix.

If you’d like to pick up another first contribution, these are both currently open, unassigned, and unclaimed:

- #207 — accept all four `--context` values in the scoring CLI and add focused CLI tests. This is the smaller code-and-tests option.
- #217 — document three currently unmapped catalog rules in `detector/CATEGORIES.md` and sync the generated copies. This is the smaller documentation option.

Please comment on whichever one you prefer before starting and take one at a time, per the newcomer guidance in `CONTRIBUTING.md`. I’d be happy to review either.

PR #319, opened 2026-09-16
fix(detector): label unsegmented-script documents Unsupported script (#241)

### LINKED ISSUE #241: Report unsegmented-script documents instead of scoring them "Too short"
`countWords()` in `detector/patterns.js` counts `\S+` runs. Chinese and Japanese have no inter-word spaces, so a long document counts as one word and hits the "Too short" gate:

```
$ node -e "
const zh='这个函数返回一个承诺，调用方不应假设句柄之后仍可重用。'.repeat(50);
const r=D.analyzeText(zh); console.log(zh.length,'chars ->',r.stats.wordCount,r.label,r.document_classification);"
1350 chars -> 1 Too short UNSCORED
```

The gate CLI then passes the file silently, even at the strictest setting:

```
$ node bin/avoid-ai-writing-gate.js --threshold 0 cjk.md; echo "exit=$?"
PASS cjk.md — 0 finding(s), threshold 0
exit=0
```

`detector/README.md` documents exit 2 for an unscannable-input error, but the gate handles only `tooLong`; `tooShort` and `Empty` pass. So a repository-wide gate reports green on documents it never scored, and the reason it gives is wrong and unactionable for a 1,350-character document.

This is distinct from the localisation question in #142 and #71 (whether the rules transfer to other languages). This is about the engine being unable to say it declined.

### Decisions needed

- Should a document the engine cannot segment fail the gate, pass with a warning, or get a distinct exit code?
- Is a new label (`Unsupported script` or `Unscorable`) better than a `stats` reason on the existing `Too short`?

### Scope

`detector/patterns.js` (detect CJK ranges before the word gate and set a label or `stats` reason), `bin/avoid-ai-writing-gate.js` (the chosen behavior), `detector/README.md` label table and exit-code list, fixtures. Run both sync scripts.

### Acceptance criteria

- A CJK document of several hundred characters is not labelled `Too short`.
- `analyzeText` exposes, in `stats` or the label, that the document was declined and why.
- Gate behavior on an unscorable file matches `detector/README.md`'s exit-code list.
- A genuinely short English document still reports `Too short`.

anujsingh-cse:
Closes #241

## Problem

`countWords()` counts `\S+` runs. Chinese and Japanese carry no inter-word spaces, so a long CJK document counted as one word and hit the `Too short` gate. The gate CLI then passed the file silently even at the strictest threshold:

```
node bin/avoid-ai-writing-gate.js --threshold 0 cjk.md
PASS cjk.md — 0 finding(s), threshold 0   # never scored
```

A repository-wide gate reported green on documents it never scored, and the `Too short` reason was wrong and unactionable for a 1,350-character document.

## Fix

Both decisions from the issue, as proposed there:

- **Label**: a new `Unsupported script` unscored label (mirroring the `tooShort`/`tooLong` shape with an `unsupportedScript: true` flag), rather than a `stats` reason on `Too short`. `stats` carries the why: `cjkChars` count and a `reason` string.
- **Detection**: Han + kana ranges checked before the word gate. Hangul is space-separated and segments fine with `\S+` runs, so Korean documents still segment and score normally.
- **Gate behavior**: `unsupportedScript` is handled like `tooLong` — exit 2 with a `cannot scan` message on stderr, matching `detector/README.md`'s documented unscannable-input exit code.

## Tests

- `patterns.test.js`: `#241: unsegmented-script documents are declined, not scored "Too short"` — the CJK repro, a Japanese document, a genuinely short English document still reporting `Too short`, and a Korean document still scoring normally
- `avoid-ai-writing-gate.test.js`: a CJK file exits 2 with `unsegmented-script document` on stderr and no `PASS` line

README label table and exit-code list updated; `## [Unreleased]` changelog entry added; bundled `patterns.js` copies regenerated via both sync scripts. `npm test` passes.


anujsingh-cse:
Fixed in the amended commit (force-pushed). All four points addressed:

1. **Halfwidth katakana** — added `U+FF66–U+FF9D` to the range. The halfwidth check now runs before the word gate, so newline-wrapped documents can't bypass it.

2. **Multiline CJK** — moved the check out of the `wordCount < 10` branch. A 10-line newline-separated Chinese document is now declined.

3. **Short mixed-language false positive** — replaced the simple `cjkChars > 0` gate with a **dominance check**: `cjkChars * 2 >= nonSpaceChars`. A short English doc with "Tokyo (東京)" is no longer declined — it still reports `Too short`.

4. **Declined files in the self-scan** — `scoreLongText` and `score` now return `declined: true` on the row and propagate through `scanFile`, the human-readable table (shows ` DECLINED`), and `--check` (fails with "could not be scored (unsupported script)"). JSON output carries `declined: true`.

All tests pass.

conorbronsdon:
Thanks @anujsingh-cse! Making a couple small fixes then will merge

